PR #10, opened 2022-06-21
IOSim's MVars

coot:
MVar's for `IOSim`

### DIFF
--- a/io-classes/src/Control/Monad/Class/MonadMVar.hs
+++ b/io-classes/src/Control/Monad/Class/MonadMVar.hs
@@ -1,0 +1,391 @@
+{-# LANGUAGE DefaultSignatures      #-}
+{-# LANGUAGE QuantifiedConstraints  #-}
+{-# LANGUAGE TypeFamilyDependencies #-}
+
+module Control.Monad.Class.MonadMVar
+  ( MonadMVar (..)
+  , MVarDefault
+  , newEmptyMVarDefault
+  , newMVarDefault
+  , putMVarDefault
+  , takeMVarDefault
+  , readMVarDefault
+  , tryTakeMVarDefault
+  , tryPutMVarDefault
+  , isEmptyMVarDefault
+  ) where
+
+import qualified Control.Concurrent.MVar as IO
+import           Control.Exception (SomeAsyncException (..))
+import           Control.Monad.Class.MonadSTM
+import           Control.Monad.Class.MonadThrow
+
+import           Control.Monad.Reader (ReaderT (..))
+import           Control.Monad.Trans (lift)
+
+import           Data.Kind (Type)
+import           Deque.Strict (Deque)
+import qualified Deque.Strict as Deque
+
+
+class Monad m => MonadMVar m where
+  {-# MINIMAL newEmptyMVar, takeMVar, putMVar, tryTakeMVar, tryPutMVar, isEmptyMVar #-}
+
+  type MVar m = (mvar :: Type -> Type) | mvar -> m
+
+  newEmptyMVar      :: m (MVar m a)
+  takeMVar          :: MVar m a -> m a
+  putMVar           :: MVar m a -> a -> m ()
+  tryTakeMVar       :: MVar m a -> m (Maybe a)
+  tryPutMVar        :: MVar m a -> a -> m Bool
+  isEmptyMVar       :: MVar m a -> m Bool
+
+  -- methods with a default implementation
+  newMVar           :: a -> m (MVar m a)
+  readMVar          :: MVar m a -> m a
+  swapMVar          :: MVar m a -> a -> m a
+  withMVar          :: MVar m a -> (a -> m b) -> m b
+  withMVarMasked    :: MVar m a -> (a -> m b) -> m b
+  modifyMVar_       :: MVar m a -> (a -> m a) -> m ()
+  modifyMVar        :: MVar m a -> (a -> m (a, b)) -> m b
+  modifyMVarMasked_ :: MVar m a -> (a -> m a) -> m ()
+  modifyMVarMasked  :: MVar m a -> (a -> m (a,b)) -> m b
+
+  default newMVar :: a -> m (MVar m a)
+  newMVar a = do
+    v <- newEmptyMVar
+    putMVar v a
+    return v
+  {-# INLINE newMVar #-}
+
+  default readMVar :: MVar m a -> m a
+  readMVar v = do
+    a <- takeMVar v
+    putMVar v a
+    return a
+  {-# INLINE readMVar #-}
+
+  default swapMVar :: MonadMask m => MVar m a -> a -> m a
+  swapMVar mvar new =
+    mask_ $ do
+      old <- takeMVar mvar
+      putMVar mvar new
+      return old
+  {-# INLINE swapMVar #-}
+
+  default withMVar :: MonadMask m => MVar m a -> (a -> m b) -> m b
+  withMVar m io =
+    mask $ \restore -> do
+      a <- takeMVar m
+      b <- restore (io a) `onException` putMVar m a
+      putMVar m a
+      return b
+  {-# INLINE withMVar #-}
+
+  default withMVarMasked :: MonadMask m => MVar m a -> (a -> m b) -> m b
+  withMVarMasked m io =
+    mask_ $ do
+      a <- takeMVar m
+      b <- io a `onException` putMVar m a
+      putMVar m a
+      return b
+  {-# INLINE withMVarMasked #-}
+
+  default modifyMVar_ :: MonadMask m => MVar m a -> (a -> m a) -> m ()
+  modifyMVar_ m io =
+    mask $ \restore -> do
+      a  <- takeMVar m
+      a' <- restore (io a) `onException` putMVar m a
+      putMVar m a'
+  {-# INLINE modifyMVar_ #-}
+
+  default modifyMVar :: (MonadMask m, MonadEvaluate m)
+                     => MVar m a -> (a -> m (a,b)) -> m b
+  modifyMVar m io =
+    mask $ \restore -> do
+      a      <- takeMVar m
+      (a',b) <- restore (io a >>= evaluate) `onException` putMVar m a
+      putMVar m a'
+      return b
+  {-# INLINE modifyMVar #-}
+
+  default modifyMVarMasked_ :: MonadMask m => MVar m a -> (a -> m a) -> m ()
+  modifyMVarMasked_ m io =
+    mask_ $ do
+      a  <- takeMVar m
+      a' <- io a `onException` putMVar m a
+      putMVar m a'
+  {-# INLINE modifyMVarMasked_ #-}
+
+  default modifyMVarMasked :: (MonadMask m, MonadEvaluate m)
+                           => MVar m a -> (a -> m (a,b)) -> m b
+  modifyMVarMasked m io =
+    mask_ $ do
+      a      <- takeMVar m
+      (a',b) <- (io a >>= evaluate) `onException` putMVar m a
+      putMVar m a'
+      return b
+  {-# INLINE modifyMVarMasked #-}
+
+
+instance MonadMVar IO where
+    type MVar IO      = IO.MVar
+    newEmptyMVar      = IO.newEmptyMVar
+    newMVar           = IO.newMVar
+    takeMVar          = IO.takeMVar
+    putMVar           = IO.putMVar
+    readMVar          = IO.readMVar
+    swapMVar          = IO.swapMVar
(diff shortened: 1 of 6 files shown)


io-classes/src/Control/Monad/Class/MonadMVar.hs
```diff
@@ -0,0 +1,379 @@
...
+              MVarFull x' blockedq -> do
+                -- async exception was thrown while we were blocked on wakevar;
+                -- we need to remove it from the queue, otherwise we will have
+                -- a space leak.
+                let blockedq' = Deque.filter ((/= wakevar) . snd) blockedq
+                writeTVar tv (MVarFull x' blockedq')
+              -- the exception was thrown when we were blocked on 'waketvar', so
+              -- the 'MVar' must not be empty.
+              MVarEmpty {} -> error "putMVarDefault: invariant violation"
+          throwIO e
```
bolt12, line 214: should this run masked? I am not sure if `catch` runs the handler masked

io-classes/src/Control/Monad/Class/MonadMVar.hs
```diff
@@ -0,0 +1,379 @@
...
+              MVarEmpty blockedq -> do
+                -- async exception was thrown while were were blocked on
+                -- wakevar; we need to remove it from 'blockedq', otherwise we
+                -- will have a space leak.
+                let blockedq' = Deque.filter (/= wakevar) blockedq
+                writeTVar tv (MVarEmpty blockedq')
+              -- the exception was thrown while we were blocked on 'wakevar', so
+              -- the 'MVar' must not be full.
+              MVarFull {} -> error "takeMVarDefault: invariant violation"
+          throwIO e
```
bolt12, line 266: Same here

io-classes/src/Control/Monad/Class/MonadMVar.hs
```diff
@@ -0,0 +1,379 @@
...
+      s <- readTVar tv
+      case s of
+        MVarEmpty _ -> return Nothing
+        MVarFull x blockedq ->
+          case Deque.uncons blockedq of
+            Nothing -> return (Just x)
+            Just ((x', wakevar), blockedq') -> do
+              writeTVar wakevar True
+              writeTVar tv (MVarFull x' blockedq')
+              return (Just x)
```
bolt12, line 304: Might be worth to generalize this to something like `takeSTM` and reuse in `takeMVarDefault` as well

io-classes/src/Control/Monad/Class/MonadMVar.hs
```diff
@@ -0,0 +1,379 @@
...
+        MVarEmpty blockedq ->
+          case Deque.uncons blockedq of
+            Nothing -> do
+              writeTVar tv (MVarFull x mempty)
+              return True
+
+            Just (wakevar, blockedq') -> do
+              writeTVar wakevar (Just x)
+              writeTVar tv (MVarEmpty blockedq')
+              return True
```
bolt12, line 324: Same for `put`

io-sim/test/Test/Control/Monad/Class/MonadMVar.hs
```diff
@@ -0,0 +1,325 @@
...
+
+
+--
+-- putMVar
+--
+
+-- | Check that 'takeMVar' is fair.  This is test is only designed for 'IOSim'
+-- as it relies on its thread scheduling and determinism.
+--
+prop_putMVar_fairness
```
bolt12, line 75: maybe drop the `prop_` prefix from this as it might be confusing. I am not sure what's the best convention

io-sim/test/Test/Control/Monad/Class/MonadMVar.hs
```diff
@@ -0,0 +1,325 @@
...
+    putMVar v ()
+    end <- getMonotonicTime
+    return (end `diffTime` start >= delta)
+
+unit_putMVar_blocks_on_full_sim :: Assertion
+unit_putMVar_blocks_on_full_sim = assertBool "did not block on an full MVar" $
+    runSimOrThrow unit_putMVar_blocks_on_full
+
+unit_putMVar_blocks_on_full_io :: Assertion
+unit_putMVar_blocks_on_full_io =
```
bolt12, line 125: Maybe add `prop_` prefix to these?

io-sim/test/Test/Control/Monad/Class/MonadMVar.hs
```diff
@@ -0,0 +1,325 @@
...
+prop_takeMVar_fairness
+  :: forall m.
+     ( MonadAsync m
+     , MonadDelay m
+     , MonadMVar  m
+     , Eq (Async m Int)
+     )
+  => Int -- ^ number of threads
+  -> m Property
+prop_takeMVar_fairness n = do
```
bolt12, line 145: Same here (drop `prop_`)

io-sim/test/Test/Control/Monad/Class/MonadMVar.hs
```diff
@@ -0,0 +1,325 @@
...
+     )
+  => Bool -> Bool
+prop_tryPutMVar isEmpty = runSimOrThrow $ do
+    v <- if isEmpty
+         then newEmptyMVar
+         else newMVar ()
+    start <- getMonotonicTime
+    _ <- tryPutMVar v ()
+    end <- getMonotonicTime
+    return $ start == end
```
bolt12, line 249: Another way would be to use `timeout` with a full `MVar` and check that it does not timeout. This way wouldn't be specific to `IOSim`

bolt12 (commented):
Looks good! Just some minor comments.

GHC Timeout API uses `MVars` in its implementation, mine https://github.com/input-output-hk/io-sim/pull/9 does not because at the time there wasn't a suitable `MVar` definition in IOSim. I don't think I will change my PR to use `MVar`s because these ones are based in STM and the whole purpose of my PR is to stay away from STM. My implementation mimics `MVar` behavior in `IOSim` so it should be good enough for now... I am not sure what @dcoutts is going to say about these `MVar`s but a truly faithful `MVar` implementation should be specific to `IOSim` and thus implemented there, right?

bolt12:
Also stylish is failing in CI

coot:
> I don't think I will change my PR to use MVars

You're using the `io-sim` `TVar`'s there, isn't it? The advantage of these `MVar`'s is that they are provide fairness, unlike `TMVar`'s.  

btw, this implementation relies on @dcoutts idea.

io-classes/src/Control/Monad/Class/MonadMVar.hs
coot: as in the `putMVarDefault` case

io-classes/src/Control/Monad/Class/MonadMVar.hs
coot: this one does not need to be masked, but yeah maybe can share some code.

io-sim/test/Test/Control/Monad/Class/MonadMVar.hs
coot: Isn't it a property?

io-sim/test/Test/Control/Monad/Class/MonadMVar.hs
coot: but how then check that `putMVar` doesn't block when the `MVar` is empty?

io-classes/src/Control/Monad/Class/MonadMVar.hs
bolt12: Oh right I did not noticed that! Sorry!

io-sim/test/Test/Control/Monad/Class/MonadMVar.hs
bolt12: Ah good point, I think the critical case is when the `MVar` is full, if it is empty and we have a test for `tryTakeMVar` not blocking, that covers some part of the behavior we want when the `MVar` is empty. Do we check `take` and `put` are non blocking for the full, empty case, respectively? Then it might be sufficient to make an argument. But this already sounds like we are trying to hard 😛

io-sim/test/Test/Control/Monad/Class/MonadMVar.hs
coot: Right, I'll change the name.

io-sim/test/Test/Control/Monad/Class/MonadMVar.hs
coot: Another way is to check if the trace does not contain any `stm` blocking operation. This would be still `IOSim` sepecific, but a nice example how one could use the low level trace.  But then the test relies on correctness of the `io-sim` tracing mechanism.

bolt12:
> You're using the io-sim TVar's there, isn't it? The advantage of these MVar's is that they are provide fairness, unlike TMVar's.

> btw, this implementation relies on @dcoutts idea.


I was, now it is using STRef's. How does STM provide fairness, I remember seeing that STM doesn't exactly guarantee fairness like MVars do

coot:
> How does STM provide fairness, I remember seeing that STM doesn't exactly guarantee fairness like MVars do

`STM` doesn't provide fairness, that's why the dance with `MVarState` and both `putMVar` and `takeMVar` are not implemented using a single atomic `STM` action.

io-sim/test/Test/Control/Monad/Class/MonadMVar.hs
```diff
@@ -0,0 +1,325 @@
...
+
+unit_tryPutMVar_empty :: Assertion
+unit_tryPutMVar_empty = assertBool "blocked on an empty MVar" $
+    prop_tryTakeMVar False
```
coot, line 253: Ought to be `prop_tryPutMVar`.

io-sim/test/Test/Control/Monad/Class/MonadMVar.hs
```diff
@@ -0,0 +1,325 @@
...
+
+unit_tryPutMVar_full :: Assertion
+unit_tryPutMVar_full = assertBool "blocked on an empty MVar" $
+    prop_tryTakeMVar True
```
coot, line 257: Ought to be `prop_tryPutMVar`.

io-sim/test/Test/Control/Monad/Class/MonadMVar.hs
coot: Fixed.

io-sim/test/Test/Control/Monad/Class/MonadMVar.hs
coot: Renamed with a suffix `_property`.

io-sim/test/Test/Control/Monad/Class/MonadMVar.hs
coot: For unitests we've been using `unit_` prefix.

io-classes/src/Control/Monad/Class/MonadMVar.hs
coot: I am not sure if it's worth it. The first atomic block of `takeMVarDefault` and `tryTakeMVar` differs in the `MVarEmpty` case, so we will still need to match on the result of `takeSTM`.  If we do that will also provide an indirection which is not that convenient when reading the code.

bolt12: Agree! 😄

bolt12:
> STM doesn't provide fairness, that's why the dance with MVarState and both putMVar and takeMVar are not implemented using a single atomic STM action.

So when you say:

> Including a default implementation using 'MonadSTM', which guarantees fairness.

what do you mean by fairness? what I understand by fairness here is that if you have various thread blocked on waiting for a `take` in a loop, and one thread putting in the MVar in a loop, eventually all threads will be able to acquire the `take`. I do not think STM guarantees this fairness

coot:
> what do you mean by fairness? 

It means blocked `putMVar`s (or `takeMVar`s) are served in FIFO order.  `STM` indeed does not guarantee that, and that's why `putMVar` and `takeMVar` are more complex than what `TMVar` would provide, also note that we need to `atomically` blocks not a single one!  Does [this](https://github.com/input-output-hk/io-sim/blob/coot/mvar/io-classes/src/Control/Monad/Class/MonadMVar.hs#L158-L166) explains it well?

coot:
The same fairness (FIFO order) is guaranteed by `MVar`'s provided by `base`: [ref](https://hackage.haskell.org/package/base-4.17.0.0/docs/GHC-MVar.html#v:takeMVar).

bolt12:
@coot  I understand now! Thank you! :D

bolt12 (approved):

PR #13, opened 2022-07-11
Add MonadCatch instance for STM

yogeshsajanikar:
WIP
- Add MonadCatch Instance
- Add CatchStm and CatchStmFrame
- Add handler for CatchStmFrame


io-sim/src/Control/Monad/IOSim/Types.hs
```diff
@@ -269,9 +272,9 @@ instance Monad (STM s) where
     {-# INLINE (>>) #-}
     (>>) = (*>)
 
-#if !(MIN_VERSION_base(4,13,0))
```
yogeshsajanikar, line 272: Reinstantiate these changes.

yogeshsajanikar:
This PR should be superceded by #15 and #16

yogeshsajanikar:
Aborting this PR in favor of #15 and #16

PR #16, opened 2022-08-02
CAD-4738 stm monad catch instance

yogeshsajanikar:
Based on the branch "CAD-4750-generalise-orelse-frame", this commit introduces `Catch` frame handler. It also adds Catch term in the reference STM for testing. 

Resolves input-output-hk/ouroboros-network#1461

### DIFF
--- a/io-sim/src/Control/Monad/IOSim/Internal.hs
+++ b/io-sim/src/Control/Monad/IOSim/Internal.hs
@@ -1148,19 +1148,45 @@ execAtomically !time !tid !tlbl !nextVid0 action0 k0 =
 
       ThrowStm e ->
         {-# SCC "execAtomically.go.ThrowStm" #-} do
-        -- Revert all the TVar writes
+        -- Rollback `TVar`s written since catch handler was installed
         !_ <- traverse_ (\(SomeTVar tvar) -> revertTVar tvar) written
-        k0 $ StmTxAborted [] (toException e)
+        case ctl of
+          AtomicallyFrame -> do
+            k0 $ StmTxAborted (Map.elems read) (toException e)
+
+          BranchFrame (CatchStmA h) k writtenOuter writtenOuterSeq createdOuterSeq ctl' ->
+            {-# SCC "execAtomically.go.BranchFrame" #-} do
+            -- Execute the left side in a new frame with an empty written set.
+            -- but preserve ones that were set prior to it, as specified in the
+            -- [stm](https://hackage.haskell.org/package/stm/docs/Control-Monad-STM.html#v:catchSTM) package.
+            let ctl'' = BranchFrame NoOpStmA k writtenOuter writtenOuterSeq createdOuterSeq ctl'
+            go ctl'' read Map.empty [] [] nextVid (h e)
+
+          BranchFrame (OrElseStmA _r) _k writtenOuter writtenOuterSeq createdOuterSeq ctl' ->
+            {-# SCC "execAtomically.go.BranchFrame" #-} do
+            go ctl' read writtenOuter writtenOuterSeq createdOuterSeq nextVid (ThrowStm e)
+
+          BranchFrame NoOpStmA _k writtenOuter writtenOuterSeq createdOuterSeq ctl' ->
+            {-# SCC "execAtomically.go.BranchFrame" #-} do
+            go ctl' read writtenOuter writtenOuterSeq createdOuterSeq nextVid (ThrowStm e)
+
+      CatchStm a h k ->
+        {-# SCC "execAtomically.go.ThrowStm" #-} do
+        -- Execute the catch handler with an empty written set.
+        -- but preserve ones that were set prior to it, as specified in the
+        -- [stm](https://hackage.haskell.org/package/stm/docs/Control-Monad-STM.html#v:catchSTM) package.
+        let ctl' = BranchFrame (CatchStmA h) k written writtenSeq createdSeq ctl
+        go ctl' read Map.empty [] [] nextVid a
+
 
       Retry ->
-        {-# SCC "execAtomically.go.Retry" #-}
-        do
+        {-# SCC "execAtomically.go.Retry" #-} do
           -- Always revert all the TVar writes for the retry
           !_ <- traverse_ (\(SomeTVar tvar) -> revertTVar tvar) written
           case ctl of
             AtomicallyFrame -> do
               -- Return vars read, so the thread can block on them
-              k0 $! StmTxBlocked $! (Map.elems read)
+              k0 $! StmTxBlocked $! Map.elems read
 
             BranchFrame (OrElseStmA b) k writtenOuter writtenOuterSeq createdOuterSeq ctl' ->
               {-# SCC "execAtomically.go.BranchFrame.OrElseStmA" #-} do
--- a/io-sim/src/Control/Monad/IOSim/Types.hs
+++ b/io-sim/src/Control/Monad/IOSim/Types.hs
@@ -195,6 +195,7 @@ runSTM (STM k) = k ReturnStm
 data StmA s a where
   ReturnStm    :: a -> StmA s a
   ThrowStm     :: SomeException -> StmA s a
+  CatchStm     :: StmA s a -> (SomeException -> StmA s a) -> (a -> StmA s b) -> StmA s b
 
   NewTVar      :: Maybe String -> x -> (TVar s x -> StmA s b) -> StmA s b
   LabelTVar    :: String -> TVar s a -> StmA s b -> StmA s b
@@ -338,6 +339,31 @@ instance MonadThrow (STM s) where
 
 instance Exceptions.MonadThrow (STM s) where
   throwM = MonadThrow.throwIO
+
+
+instance MonadCatch (STM s) where
+
+  catch action handler = STM $ oneShot $ \k -> CatchStm (runSTM action) (runSTM . fromHandler handler) k
+    where
+      -- Get a total handler from the given handler
+      fromHandler :: Exception e => (e -> STM s a) -> SomeException -> STM s a
+      fromHandler h e = case fromException e of
+        Nothing -> throwIO e  -- Rethrow the exception if handler does not handle it.
+        Just e' -> h e'
+
+  -- Masking is not required as STM actions are always run inside
+  -- `execAtomically` and behave as if masked. Also note that the default
+  -- implementation of `generalBracket` needs mask, and is part of `MonadThrow`.
+  generalBracket acquire release use = do
+    resource <- acquire
+    b <- use resource `catch` \e -> do
+      _ <- release resource (ExitCaseException e)
+      throwIO e
+    c <- release resource (ExitCaseSuccess b)
+    return (b, c)
+
+instance Exceptions.MonadCatch (STM s) where
+  catch = MonadThrow.catch
 
 instance MonadCatch (IOSim s) where
   catch action handler =
@@ -867,9 +893,22 @@ data StmTxResult s a =
      | StmTxAborted  [SomeTVar s] SomeException
 
 
--- | OrElse/Catch give rise to an alternate right hand side branch. A right branch
--- can be a NoOp
-data BranchStmA s a = OrElseStmA (StmA s a) | NoOpStmA
+-- | A branch indicates that an alternative statement is available in the current
+-- context. For example, `OrElse` has two alternative statements, say "left"
+-- and "right". While executing the left statement, `OrElseStmA` branch indicates
+-- that the right branch is still available, in case the left statement fails.
+data BranchStmA s a =
+       -- | `OrElse` statement with its 'right' alternative.
+       OrElseStmA (StmA s a)
+       -- | `CatchStm` statement with the 'catch' handler.
+     | CatchStmA (SomeException -> StmA s a)
+       -- | Unlike the other two branches, the no-op branch is not an explicit
+       -- part of the STM syntax. It simply indicates that there are no
+       -- alternative statements left to be executed. For example, when running
+       -- right alternative of the `OrElse` statement or when running the catch
+       -- handler of a `CatchStm` statement, there are no alternative statements
+       -- available. This case is represented by the no-op branch.
+     | NoOpStmA
 
 data StmStack s b a where
   -- | Executing in the context of a top level 'atomically'.
--- a/io-sim/src/Control/Monad/IOSimPOR/Internal.hs
+++ b/io-sim/src/Control/Monad/IOSimPOR/Internal.hs
@@ -1391,32 +1391,54 @@ execAtomically time tid tlbl nextVid0 action0 k0 =
         {-# SCC "execAtomically.go.ThrowStm" #-} do
         -- Revert all the TVar writes
         !_ <- traverse_ (\(SomeTVar tvar) -> revertTVar tvar) written
-        k0 $ StmTxAborted (Map.elems read) (toException e)
+        case ctl of
+          AtomicallyFrame -> do
+            k0 $ StmTxAborted (Map.elems read) (toException e)
+
+          BranchFrame (CatchStmA h) k writtenOuter writtenOuterSeq createdOuterSeq ctl' ->
+            {-# SCC "execAtomically.go.BranchFrame" #-} do
+            -- Execute the left side in a new frame with an empty written set.
+            -- but preserve ones that were set prior to it, as specified in the
+            -- [stm](https://hackage.haskell.org/package/stm/docs/Control-Monad-STM.html#v:catchSTM) package.
+            let ctl'' = BranchFrame NoOpStmA k writtenOuter writtenOuterSeq createdOuterSeq ctl'
+            go ctl'' read Map.empty [] [] nextVid (h e)
+
+          BranchFrame (OrElseStmA _r) _k writtenOuter writtenOuterSeq createdOuterSeq ctl' ->
+            {-# SCC "execAtomically.go.BranchFrame" #-} do
+            go ctl' read writtenOuter writtenOuterSeq createdOuterSeq nextVid (ThrowStm e)
+
+          BranchFrame NoOpStmA _k writtenOuter writtenOuterSeq createdOuterSeq ctl' ->
+            {-# SCC "execAtomically.go.BranchFrame" #-} do
+            go ctl' read writtenOuter writtenOuterSeq createdOuterSeq nextVid (ThrowStm e)
+
+      CatchStm a h k ->
+        {-# SCC "execAtomically.go.ThrowStm" #-} do
+        -- Execute the left side in a new frame with an empty written set
+        let ctl' = BranchFrame (CatchStmA h) k written writtenSeq createdSeq ctl
+        go ctl' read Map.empty [] [] nextVid a
 
       Retry ->
-        {-# SCC "execAtomically.go.Retry" #-}
-        do
-          -- Always revert all the TVar writes for the retry
-          !_ <- traverse_ (\(SomeTVar tvar) -> revertTVar tvar) written
-          case ctl of
-            AtomicallyFrame -> do
-              -- Return vars read, so the thread can block on them
-              k0 $! StmTxBlocked $! Map.elems read
+        {-# SCC "execAtomically.go.Retry" #-} do
+        -- Always revert all the TVar writes for the retry
+        !_ <- traverse_ (\(SomeTVar tvar) -> revertTVar tvar) written
+        case ctl of
+          AtomicallyFrame -> do
+            -- Return vars read, so the thread can block on them
+            k0 $! StmTxBlocked $! Map.elems read
 
-            BranchFrame (OrElseStmA b) k writtenOuter writtenOuterSeq createdOuterSeq ctl' ->
-              {-# SCC "execAtomically.go.BranchFrame.OrElseStmA" #-} do
-              !_ <- traverse_ (\(SomeTVar tvar) -> revertTVar tvar) written
-              -- Execute the orElse right hand with an empty written set
-              let ctl'' = BranchFrame NoOpStmA k writtenOuter writtenOuterSeq createdOuterSeq ctl'
-              go ctl'' read Map.empty [] [] nextVid b
+          BranchFrame (OrElseStmA b) k writtenOuter writtenOuterSeq createdOuterSeq ctl' ->
+            {-# SCC "execAtomically.go.BranchFrame.OrElseStmA" #-} do
+            -- Execute the orElse right hand with an empty written set
+            let ctl'' = BranchFrame NoOpStmA k writtenOuter writtenOuterSeq createdOuterSeq ctl'
+            go ctl'' read Map.empty [] [] nextVid b
 
-            BranchFrame _ _k writtenOuter writtenOuterSeq createdOuterSeq ctl' ->
-              {-# SCC "execAtomically.go.BranchFrame" #-} do
-              -- Retry makes sense only within a OrElse context. If it is a branch other than
-              -- OrElse left side, then bubble up the `retry` to the frame above.
-              -- Skip the continuation and propagate the retry into the outer frame
-              -- using the written set for the outer frame
-              go ctl' read writtenOuter writtenOuterSeq createdOuterSeq nextVid Retry
+          BranchFrame _ _k writtenOuter writtenOuterSeq createdOuterSeq ctl' ->
+            {-# SCC "execAtomically.go.BranchFrame" #-} do
+            -- Retry makes sense only within a OrElse context. If it is a branch other than
+            -- OrElse left side, then bubble up the `retry` to the frame above.
+            -- Skip the continuation and propagate the retry into the outer frame
+            -- using the written set for the outer frame
+            go ctl' read writtenOuter writtenOuterSeq createdOuterSeq nextVid Retry
 
       OrElse a b k ->
         {-# SCC "execAtomically.go.OrElse" #-} do
--- a/io-sim/test/Test/IOSim.hs
+++ b/io-sim/test/Test/IOSim.hs
@@ -1249,7 +1249,7 @@ prop_stm_referenceSim t =
 -- | Compare the behaviour of the STM reference operational semantics with
 -- the behaviour of any 'MonadSTM' STM implementation.
 --
-prop_stm_referenceM :: (MonadSTM m, MonadThrow (STM m), MonadCatch m)
+prop_stm_referenceM :: (MonadSTM m, MonadCatch (STM m), MonadCatch m)
                     => SomeTerm -> m Property
 prop_stm_referenceM (SomeTerm _tyrep t) = do
     let (r1, _heap) = evalAtomically t
--- a/io-sim/test/Test/STM.hs
+++ b/io-sim/test/Test/STM.hs
@@ -67,6 +67,7 @@ data Term (t :: Type) where
 
     Return    :: Expr t -> Term t
     Throw     :: Expr a -> Term t
+    Catch     :: Term t -> Term t -> Term t
     Retry     :: Term t
 
     ReadTVar  :: Name (TyVar t) -> Term t
@@ -296,7 +297,7 @@ deriving instance Show (NfTerm t)
 -- | The STM transition rules. They reduce a 'Term' to a normal-form 'NfTerm'.
 --
 -- Compare the implementation of this against the operational semantics in
--- Figure 4 in the paper. Note that @catch@ is not included.
+-- Figure 4 in the paper including the `Catch` semantics from the Appendix A.
 --
 evalTerm :: Env -> Heap -> Allocs -> Term t -> (NfTerm t, Heap, Allocs)
 evalTerm !env !heap !allocs term = case term of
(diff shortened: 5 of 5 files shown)


io-sim/src/Control/Monad/IOSim/Internal.hs
```diff
@@ -896,10 +896,33 @@ execAtomically !time !tid !tlbl !nextVid0 action0 k0 =
           go ctl' read written' writtenSeq' createdOuterSeq nextVid (k x)
 
       ThrowStm e ->
+        {-# SCC "execAtomically.go.ThrowStm" #-}
+        case ctl of 
+          AtomicallyFrame -> do
+            -- Revert all the TVar writes
+            !_ <- traverse_ (\(SomeTVar tvar) -> revertTVar tvar) written
```
amesgen, line 903: This is shared between all branches, maybe do it before the case branch?

io-sim/test/Test/IOSim.hs
```diff
@@ -1049,7 +1049,7 @@ prop_stm_referenceSim t =
 -- | Compare the behaviour of the STM reference operational semantics with
 -- the behaviour of any 'MonadSTM' STM implementation.
 --
-prop_stm_referenceM :: (MonadSTM m, MonadThrow (STM m), MonadCatch m)
+prop_stm_referenceM :: (MonadSTM m, MonadThrow (STM m), MonadCatch m, LazySTM.MonadSTM m, MonadCatch (LazySTM.STM m))
```
amesgen, line 1052: Small simplification:
```suggestion
prop_stm_referenceM :: (MonadSTM m, MonadCatch (STM m), MonadCatch m)
```

io-sim/test/Test/STM.hs
```diff
@@ -438,7 +462,7 @@ extendExecEnv (Name name _tyrep) v (ExecEnv env) =
 
 -- | Execute an STM 'Term' in the 'STM' monad.
 --
-execTerm :: (MonadSTM m, MonadThrow (STM m))
+execTerm :: (MonadSTM m, MonadThrow (STM m), MonadCatch (STM m))
```
amesgen, line 465:
```suggestion
execTerm :: (MonadSTM m, MonadCatch (STM m))
```

io-sim/test/Test/STM.hs
```diff
@@ -492,7 +518,7 @@ snapshotExecValue (ExecValInt x)   = return (ImmValInt x)
 snapshotExecValue (ExecValVar v _) = fmap ImmValVar
                                           (snapshotExecValue =<< readTVar v)
 
-execAtomically :: forall m t. (MonadSTM m, MonadThrow (STM m), MonadCatch m)
+execAtomically :: forall m t. (MonadSTM m, MonadThrow (STM m), MonadCatch m, MonadCatch (STM m))
```
amesgen, line 521:
```suggestion
execAtomically :: forall m t. (MonadSTM m, MonadCatch (STM m), MonadCatch m)
```

io-sim/test/Test/STM.hs
```diff
@@ -676,6 +702,11 @@ genTerm env tyrep =
         OrElse <$> genTerm env tyrep
                <*> genTerm env tyrep
 
+    catchTerm =
+      sized $ \sz -> resize (sz `div` 2) $
```
amesgen, line 706:
```suggestion
      scale (`div` 2) $
```
(feel free to ignore if you think it is not worth to change `orElseTerm` in the same way)

io-sim/test/Test/STM.hs
```diff
@@ -310,6 +310,30 @@ evalTerm !env !heap !allocs term = case term of
...
+        --                M; heap, {} => return P; heap', allocs'
+        -- --------------------------------------------------------
+        -- S[catch M N]; heap, allocs => S[return P]; heap', allocs'
+        NfReturn v -> (NfReturn v, heap', allocs `mappend` allocs')
+
+        -- Rule XSTM2
+        --                M; heap, {} => throw P; heap', allocs'
+        -- --------------------------------------------------------
+        -- S[catch M N]; heap, allocs => S[N P]; heap U allocs', allocs U allocs'
+        NfThrow _ -> evalTerm env (heap `mappend` allocs') (allocs `mappend` allocs') t2
```
amesgen, line 328: Usually (also in this module), people prefer `<>` (from `Semigroup`) over `mappend` (from `Monoid`) as it is both more general and nicer to read as a symbolic infix operator.

amesgen (commented):
Looks great, always cool to see when things get simpler over time :tada:

I only have a few minor comments below. As before, I did not commented on the `IOSimPOR` part as it is identical.

io-sim/test/Test/STM.hs
```diff
@@ -451,6 +476,8 @@ execTerm env t =
         let e' = execExpr env e
         throwSTM =<< snapshotExecValue e'
 
+      Catch t1 t2 -> execTerm env t1 `catch` \(_ :: ImmValue) -> execTerm env t2
```
nfrisby, line 479: Yogesh pointed out that this only tests catch's with total handlers. The terms we generate currently do not have a way to express values of different types, and so we cannot generate a partial handler. That means these tests are somewhat incomplete.

nfrisby: One option: write a separate test dedicated to checking that partial handlers work as expected.

nfrisby: And then there's the additional gap where the exception value isn't being used in the handler. I had completely overlooked that, just focused on the type-level totality :)

That'd be much more complicated to fixup, I assume, since it would require generating a well-formed lambda.

Another possible compromise: add polymorphic constant function, eg `isZero` which is `False` for `()`, `2`, etc but `True` for `isZero 0`. .... hmm, though I realize `Term` doesn't even have lexical variable occurrences. So I doubt this would be small enough to include in this PR.

nfrisby (changes requested):
I'll request some changes, just since I hope Yogesh finds a sufficiently lightweight way to address a couple of the eg testing comments. Nice stuff!

io-sim/test/Test/STM.hs
```diff
@@ -713,6 +745,9 @@ shrinkTerm t =
     case t of
       Return e      -> [Return e' | e' <- shrinkExpr e]
       Throw e       -> [Throw  e' | e' <- shrinkExpr e]
+      Catch t1 t2   -> [t1, t2]
```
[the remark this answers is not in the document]
amesgen: In `Arbitrary2 (,)`, we have
```haskell
  liftShrink2 shrA shrB (x, y) =
       [ (x', y) | x' <- shrA x ]
    ++ [ (x, y') | y' <- shrB y ]
```
so to preserve semantics, we would have to prepend `[t1,t2]`.

coot (commented):
It's on my todo list to review the PR... However, could you update `ouroboros-network` and check if all nightly tests are passing: for that one needs to add this snippet to `cabal.project.local` file:

```
package ouroboros-network-testing
  flags: -nightly
```

And then run:
```
cabal run ouroboros-netowrk-framework:test && cabal run ouroboros-network:test
```

yogeshsajanikar:
@coot  Running the nightly tests now. I will update the results here soon.

yogeshsajanikar:
@coot The tests that you mentioned pass!

Running tests in ouroboros-network with io-sim commit id 9d2e67fcb0ba066dba4803f3998b4ed2c98e1ada

> cabal run ouroboros-network-framework:test -> All 47 tests passed (118.51s)
> cabal run ouroboros-network:test -> All 314 tests passed (480.21s)

io-sim/src/Control/Monad/IOSimPOR/Internal.hs
```diff
@@ -1103,29 +1103,36 @@ execAtomically time tid tlbl nextVid0 action0 k0 =
...
-          -- Continue with the k continuation
-          go ctl' read written' writtenSeq' createdSeq' nextVid (k x)
-
       ThrowStm e ->
+        {-# SCC "execAtomically.go.ThrowStm" #-}
+        case ctl of
+          AtomicallyFrame -> do
+            -- Revert all the TVar writes
+            !_ <- traverse_ (\(SomeTVar tvar) -> revertTVar tvar) written
+            k0 $ StmTxAborted [] (toException e)
```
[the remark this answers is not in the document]
coot: Definitely a bug. `IOSimPOR` needs to compute how the clock advances by looking at the read `TVar`'s clocks.

io-sim/src/Control/Monad/IOSim/Internal.hs
```diff
@@ -892,29 +892,37 @@ execAtomically !time !tid !tlbl !nextVid0 action0 k0 =
...
+            -- Revert all the TVar writes within this branch
+            !_ <- traverse_ (\(SomeTVar tvar) -> revertTVar tvar) written
+            go ctl' read writtenOuter writtenOuterSeq createdOuterSeq nextVid (ThrowStm e)
+
+      CatchStm a h k ->
         {-# SCC "execAtomically.go.ThrowStm" #-} do
-        -- Revert all the TVar writes
-        !_ <- traverse_ (\(SomeTVar tvar) -> revertTVar tvar) written
-        k0 $ StmTxAborted [] (toException e)
+        -- Execute the left side in a new frame with an empty written set
```
coot, line 922: I would add a comment:
```suggestion
-- Execute the left side in a new frame with an empty written set.
-- Rollback `TVar`s written since catch handler was installed,
-- but preserve ones that were set prior to it, as specified in the
-- [stm](https://hackage.haskell.org/package/stm/docs/Control-Monad-STM.html#v:catchSTM) package.
```

io-sim/test/Test/STM.hs
```diff
@@ -738,6 +773,7 @@ shrinkExpr (ExprName (Name _ (TyRepVar _))) = []
 freeNamesTerm :: Term t -> Set NameId
 freeNamesTerm (Return e)      = freeNamesExpr e
 freeNamesTerm (Throw  e)      = freeNamesExpr e
+freeNamesTerm (Catch t1 t2)   = freeNamesTerm t1 <> freeNamesTerm t2
```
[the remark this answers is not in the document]
coot: `s/write now/right now` :grin:

io-sim/test/Test/STM.hs
```diff
@@ -309,6 +310,30 @@ evalTerm !env !heap !allocs term = case term of
       where
         e'  = evalExpr env e
 
+    -- Exception semantics are detailed in "Appendix A Exception semantics" p 12-13 of
```
coot, line 313: The haddock comment associated with `evalTerm` needs to be updated:
```
-- Compare the implementation of this against the operational semantics in
-- Figure 4 in the paper including `Catch` semantics from the appendix A.
```

io-sim/src/Control/Monad/IOSim/Internal.hs
```diff
@@ -892,29 +892,37 @@ execAtomically !time !tid !tlbl !nextVid0 action0 k0 =
...
-          -- Continue with the k continuation
-          go ctl' read written' writtenSeq' createdSeq' nextVid (k x)
-
       ThrowStm e ->
+        {-# SCC "execAtomically.go.ThrowStm" #-}
+        case ctl of
+          AtomicallyFrame -> do
+            -- Revert all the TVar writes
+            !_ <- traverse_ (\(SomeTVar tvar) -> revertTVar tvar) written
+            k0 $ StmTxAborted [] (toException e)
```
coot, line 904:
```suggestion
            k0 $ StmTxAborted read (toException e)
```
Although `IOSim` ignores the `read` `TVar`'s, it's better to pass them.

coot (changes requested):
Very nice, just a few remarks.

io-sim/src/Control/Monad/IOSim/Types.hs
```diff
@@ -313,6 +315,25 @@ instance MonadThrow (STM s) where
 instance Exceptions.MonadThrow (STM s) where
   throwM = MonadThrow.throwIO
 
+instance MonadCatch (STM s) where
+
+  catch action handler = STM $ oneShot $ \k -> CatchStm (runSTM action) (runSTM . handler') k
+    where
+      handler' e = case fromException e of
```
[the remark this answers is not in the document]
yogeshsajanikar: Done.

io-sim/src/Control/Monad/IOSimPOR/Internal.hs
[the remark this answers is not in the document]
yogeshsajanikar: Fixed in both IOSimPOR and IOSim (IOSim ignored it by default)

io-sim/src/Control/Monad/IOSim/Internal.hs
coot: Could you add the comment.

io-sim/src/Control/Monad/IOSim/Internal.hs
coot: That's a good idea.

io-sim/src/Control/Monad/IOSim/Types.hs
```diff
@@ -313,6 +315,25 @@ instance MonadThrow (STM s) where
...
+        Nothing -> throwIO e  -- Rethrow the exception if handler does not handle it.
+        Just e' -> handler e'
+
+  generalBracket acquire release use = do
```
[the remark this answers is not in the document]
coot: That's a good idea.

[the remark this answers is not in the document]
coot: I actually don't understand why we don't need to mask exceptions.

io-sim/src/Control/Monad/IOSim/Types.hs
```diff
@@ -313,6 +315,25 @@ instance MonadThrow (STM s) where
...
+    resource <- acquire
+    b <- use resource `catch` \e -> do
+      _ <- release resource (ExitCaseException e)
+      throwIO e
+    c <- release resource (ExitCaseSuccess b)
```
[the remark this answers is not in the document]
coot: It's like the `IO` case, which does not abort, isn't it? https://hackage.haskell.org/package/exceptions-0.10.5/docs/src/Control.Monad.Catch.html#line-338

io-sim/src/Control/Monad/IOSimPOR/Internal.hs
```diff
@@ -1103,29 +1103,36 @@ execAtomically time tid tlbl nextVid0 action0 k0 =
...
+
+          BranchFrame (CatchStmA h) k writtenOuter writtenOuterSeq createdOuterSeq ctl' ->
+            {-# SCC "execAtomically.go.BranchFrame" #-} do
+            -- Revert all the TVar writes within this catch
+            !_ <- traverse_ (\(SomeTVar tvar) -> revertTVar tvar) written
+            -- Execute the catch handler with an empty written set
+            let ctl'' = BranchFrame EmptyStmA k writtenOuter writtenOuterSeq createdOuterSeq ctl'
+            go ctl'' read Map.empty [] [] nextVid (h e)
+            --
```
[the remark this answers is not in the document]
coot: @yogeshsajanikar could you remove it.

io-sim/test/Test/STM.hs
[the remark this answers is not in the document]
coot: @yogeshsajanikar this should be addressed.

io-sim/test/Test/STM.hs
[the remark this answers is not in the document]
coot: I think this deserves a comment.

io-sim/src/Control/Monad/IOSimPOR/Internal.hs
```diff
@@ -1110,10 +1110,31 @@ execAtomically time tid tlbl nextVid0 action0 k0 =
           go ctl' read written' writtenSeq' createdSeq' nextVid (k x)
 
       ThrowStm e ->
+        {-# SCC "execAtomically.go.ThrowStm" #-}
+        case ctl of
+          AtomicallyFrame -> do
+            -- Revert all the TVar writes
+            !_ <- traverse_ (\(SomeTVar tvar) -> revertTVar tvar) written
```
coot, line 1117: As it was pointed out in the `IOSim` case, this is common in all the branches and thus case be put in front of the `case`.

io-sim/test/Test/STM.hs
yogeshsajanikar: This is not addressed in this PR

coot: Could you please create an issue for this in this repo then?

io-sim/src/Control/Monad/IOSim/Types.hs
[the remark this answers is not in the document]
yogeshsajanikar: Yes, it is similar.

io-sim/src/Control/Monad/IOSim/Types.hs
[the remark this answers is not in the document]
yogeshsajanikar: My thinkin is that, the STM do not need to worry about masking as the the operations are atomic, always run within `execAtomic`. Also `catchSTM` in `exceptions` does not see any asynchronous exceptions. Also another point to note here is that the need for masking arises from the fact that here in io-sim the function `generalBracket` needs mask, and is part of `MonadThrow`. And for STM it probably does not make much sense.

io-sim/src/Control/Monad/IOSim/Internal.hs
yogeshsajanikar: Done for both IOSim/Internal.hs and IOSimPOR/Internal.hs

io-sim/test/Test/STM.hs
yogeshsajanikar: New issue created https://github.com/input-output-hk/io-sim/issues/36. I also started implmenting it.

io-sim/src/Control/Monad/IOSim/Types.hs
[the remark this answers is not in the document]
coot: You're right, for our implementation of `STM` we deliver async exceptions outside of `execAtomically` and thus we don't need to mask exceptions.

io-sim/src/Control/Monad/IOSim/Types.hs
```diff
@@ -322,6 +323,32 @@ instance MonadThrow (STM s) where
...
+      -- Get a total handler from the given handler
+      fromHandler :: Exception e => (e -> STM s a) -> SomeException -> STM s a
+      fromHandler h e = case fromException e of
+        Nothing -> throwIO e  -- Rethrow the exception if handler does not handle it.
+        Just e' -> h e'
+
+  -- STM actions are always run inside `execAtomically` and behave as if masked
+  -- Another point to note that the default implementation of `generalBracket` needs
+  -- mask, and is part of `MonadThrow`. For STM, it probably does not make sense to have
+  -- masking.
```
coot, line 340:
```suggestion
  -- mask, and is part of `MonadThrow`. For STM, we don't need masking because
  -- async exceptions are handled outside of `execAtomically`.
```

io-sim/test/Test/STM.hs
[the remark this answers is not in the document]
coot: This wasn't addressed.

io-sim/test/Test/STM.hs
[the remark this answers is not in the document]
coot: Could you just make sure that the comment fits into `80` columns?

io-sim/test/Test/STM.hs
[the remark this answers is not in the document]
yogeshsajanikar: @coot This is addressed. We have to prepend [t1, t2] as liftShrink2 does not have [t1,t2]. Please see next line.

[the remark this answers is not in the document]
coot: Right, sorry I missed that :)

coot (approved):

io-sim/test/Test/STM.hs
```diff
@@ -739,6 +771,12 @@ shrinkExpr (ExprName (Name _ (TyRepVar _))) = []
 freeNamesTerm :: Term t -> Set NameId
 freeNamesTerm (Return e)      = freeNamesExpr e
 freeNamesTerm (Throw  e)      = freeNamesExpr e
+-- A catch handler should actually have an argument, and then the implementation
+-- should handle it. But since current implementation of catch never binds the
```
nfrisby, line 775: This sentence scared me! `But since current implementation of catch never binds the variable`

The implementation of `catch` does that, but the generator used in this test only ever generates handlers that ignore the variable.

So let's reword this sentence to clarify that it's the generation of the handler argument that is degenerate, not _the implementation_ 👍

nfrisby (changes requested):
One last round of minor comment adjustments! 👍

io-sim/src/Control/Monad/IOSim/Internal.hs
```diff
@@ -926,19 +926,42 @@ execAtomically !time !tid !tlbl !nextVid0 action0 k0 =
...
         !_ <- traverse_ (\(SomeTVar tvar) -> revertTVar tvar) written
-        k0 $ StmTxAborted [] (toException e)
+        case ctl of
+          AtomicallyFrame -> do
+            k0 $ StmTxAborted (Map.elems read) (toException e)
+
+          BranchFrame (CatchStmA h) k writtenOuter writtenOuterSeq createdOuterSeq ctl' ->
+            {-# SCC "execAtomically.go.BranchFrame" #-} do
+            -- Execute the left side in a new frame with an empty written set.
+            -- Rollback `TVar`s written since catch handler was installed,
```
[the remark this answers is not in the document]
yogeshsajanikar: Corrected in both `IOSim/Internal.hs` and `IOSimPOR/Internal.hs`

io-sim/src/Control/Monad/IOSim/Types.hs
```diff
@@ -334,6 +335,32 @@ instance MonadThrow (STM s) where
...
+  catch action handler = STM $ oneShot $ \k -> CatchStm (runSTM action) (runSTM . fromHandler handler) k
+    where
+      -- Get a total handler from the given handler
+      fromHandler :: Exception e => (e -> STM s a) -> SomeException -> STM s a
+      fromHandler h e = case fromException e of
+        Nothing -> throwIO e  -- Rethrow the exception if handler does not handle it.
+        Just e' -> h e'
+
+  -- STM actions are always run inside `execAtomically` and behave as if masked
+  -- Another point to note that the default implementation of `generalBracket` needs
```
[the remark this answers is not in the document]
yogeshsajanikar: Reworded!

io-sim/src/Control/Monad/IOSim/Types.hs
```diff
@@ -334,6 +335,32 @@ instance MonadThrow (STM s) where
...
+  -- mask, and is part of `MonadThrow`. For STM, we don't need masking because
```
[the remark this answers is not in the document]
yogeshsajanikar: Last and first statements are combined together to create a concise comment for masking behavior for STM

io-sim/test/Test/STM.hs
yogeshsajanikar: Added a description about limitation of catch term generator.

io-sim/src/Control/Monad/IOSim/Types.hs
```diff
@@ -869,9 +896,10 @@ data StmTxResult s a =
      | StmTxAborted  [SomeTVar s] SomeException
 
 
--- | OrElse/Catch give rise to an alternate right hand side branch. A right branch
--- can be a NoOp
-data BranchStmA s a = OrElseStmA (StmA s a) | NoOpStmA
+-- | A branch is an alternative of a `OrElse` or a `CatchStm` statement
```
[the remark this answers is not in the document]
yogeshsajanikar: Added documentation for `BranchStmA`

io-sim/src/Control/Monad/IOSim/Types.hs
```diff
@@ -870,9 +896,22 @@ data StmTxResult s a =
      | StmTxAborted  [SomeTVar s] SomeException
 
 
--- | OrElse/Catch give rise to an alternate right hand side branch. A right branch
--- can be a NoOp
-data BranchStmA s a = OrElseStmA (StmA s a) | NoOpStmA
+-- | A branch indicates an alternative statment available in the current context.
```
nfrisby, line 899: `statment` -> `statement`

io-sim/src/Control/Monad/IOSim/Types.hs
```diff
@@ -870,9 +896,22 @@ data StmTxResult s a =
...
+-- For example, `OrElse` has two alternative statements, say "left" and "right".
+-- While executing the left statement, `OrElseStmA` branch indicates that the
+-- right branch is still available, in case, left statement fails.
```
nfrisby, line 902: `in case, left statement fails` -> `in case the left statement fails`

io-sim/src/Control/Monad/IOSim/Types.hs
```diff
@@ -870,9 +896,22 @@ data StmTxResult s a =
...
+data BranchStmA s a =
+       -- | `OrElse` statement with its 'right' alternative.
+       OrElseStmA (StmA s a)
+       -- | `CatchStm` statment with 'catch' handler.
+     | CatchStmA (SomeException -> StmA s a)
+       -- | No op statment is not a part of the STM semantics. It simply
```
nfrisby, line 908: `No op statment is not a part of the STM semantics.` -> `Unlike the other two branches, the no-op branch is not an explicit part of the STM syntax.`

io-sim/src/Control/Monad/IOSim/Types.hs
```diff
@@ -870,9 +896,22 @@ data StmTxResult s a =
...
+       -- indicates that there are no alternative statements left to be executed.
+       -- For example, when running right alternative of the `OrElse` statement
+       -- or when running the catch handler of a `CatchStm` statement, there are
+       -- alternative statements available. This case is represented by no-op
```
nfrisby, line 912: `alternative` -> `no alternative`

This typo is important to fix.

nfrisby, line 912: `by no-op` -> `by the no-op`

io-sim/test/Test/STM.hs
```diff
@@ -67,6 +67,7 @@ data Term (t :: Type) where
 
     Return    :: Expr t -> Term t
     Throw     :: Expr a -> Term t
+    Catch     :: Term t -> Expr a -> Term t -> Term t
```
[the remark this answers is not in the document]
nfrisby: @yogeshsajanikar did you address this one?

nfrisby (changes requested):
I'm Requesting Changes one last time! I only caught spelling/grammar this time, but one of the typos is important.

Code looks great!

io-sim/test/Test/STM.hs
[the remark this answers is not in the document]
nfrisby: Yogesh is going to squash, so this will be resolved then.

io-sim/src/Control/Monad/IOSim/STM.hs
```diff
@@ -171,12 +171,10 @@ writeTBQueueDefault (TBQueue queue _size) a = do
 
 isEmptyTBQueueDefault :: MonadSTM m => TBQueueDefault m a -> STM m Bool
 isEmptyTBQueueDefault (TBQueue queue _size) = do
-  (xs, _, ys, _) <- readTVar queue
+  (xs, _, _, _) <- readTVar queue
```
nfrisby, line 174: It looks like your squash/rebase/etc is accidentally undoing PR #44, which merged 16 hours ago.

There was another merged about 4 hours ago. And then 3 days ago.

So you may want to double-check your new commit doesn't contain any other accidental changes

nfrisby (changes requested):
All the comments have been addressed 👍 

Except for the new one I'm opening right now---looks like this commit undoes a recently merged PR. That's why I'm Requesting Changes one last time.

io-sim/src/Control/Monad/IOSim/STM.hs
nfrisby: This new commit still includes this change 
image

io-sim/src/Control/Monad/IOSimPOR/Internal.hs
```diff
@@ -1391,32 +1391,54 @@ execAtomically time tid tlbl nextVid0 action0 k0 =
...
+            -- Return vars read, so the thread can block on them
+            k0 $! StmTxBlocked $! Map.elems read
+
+          BranchFrame (OrElseStmA b) k writtenOuter writtenOuterSeq createdOuterSeq ctl' ->
+            {-# SCC "execAtomically.go.BranchFrame.OrElseStmA" #-} do
+            -- Execute the orElse right hand with an empty written set
+            let ctl'' = BranchFrame NoOpStmA k writtenOuter writtenOuterSeq createdOuterSeq ctl'
+            go ctl'' read Map.empty [] [] nextVid b
+
+          BranchFrame _ _k writtenOuter writtenOuterSeq createdOuterSeq ctl' ->
```
yogeshsajanikar, line 1435: Probably it makes sense to explicitly use `(CatchStmA _ )` and `(NoOpStmA)` here, just like we did for `ThrowStm`. @nfrisby

nfrisby (approved):
The diff looks minimal again. 🙌 

Approved! Nice work 👍

io-sim/src/Control/Monad/IOSimPOR/Internal.hs
nfrisby: I agree, but I won't Request Changes over it. Feel free to change it before you merge, if you like.

io-sim/src/Control/Monad/IOSim/STM.hs
nfrisby: Fixed! 👍

PR #17, opened 2022-08-09
Use github-actions

coot:
- Added github-actions workflow
- Updated cabal files: added tested-with


bolt12 (approved):
Looks good, is there any way to disable hydra so it stops showing as failed?

PR #25, opened 2022-09-24
Add issue templates

coot:


james-iohk:
Consider issues forms instead of templates, stricter and offers mandatory fields (templates can be altered or overrided)
See example https://github.com/input-output-hk/cardano-js-sdk/blob/master/.github/ISSUE_TEMPLATE/bug.yml

coot:
@james-iohk hmm, maybe you're right.  I initially though that one can only write text in the boxes but it actually starts a markdown editor (this is indispensable when submitting logs).

coot:
@james-iohk we decided to go with templates, as they are more elastic (the writer decides how the issue will be written).  Templates are more guidelines than a strict rule, I think this will fit better the niche we are in.

bolt12 (approved):
LGTM!

PR #54, opened 2022-11-29
Cancellable Timers & Hackage Preparations

coot:


coot:
@dcoutts we have to keep `MonadTimerFancy` in `io-classes`, because we relay on it to implement `registerDelay`, but I moved it to `MonadTimer.NonStandard` module.

`registerDelayCancellable` is just a function on top of `MonadTimerFancy`, I think there's no real benefit in packaging it independently, so it's exported from `MonadTimer.NonStandar`.

PR #65, opened 2023-01-19
Fix index-state syntax

andreabedini:
The second index-state stanza completely ovverides the first, resetting hackage index state to HEAD. See haskell/cabal#8568


coot (approved):
LGTM :)

PR #73, opened 2023-03-14
Add ghc 9.6.1 to the matrix

angerman:


bolt12 (approved):

io-sim/src/Data/Deque/Strict.hs
```diff
@@ -0,0 +1,41 @@
...
+
+import           Prelude hiding (head, init, tail)
+import           Data.Foldable (foldl')
+import qualified Data.List as List
+
+data Deque a = Deque ![a] ![a]
+
+instance Semigroup (Deque a) where
+    Deque as bs <> Deque as' bs' =
+      Deque as (bs ++ reverse as' ++ bs')
```
bolt12, line 15: Shouldn't this be

```suggestion
      Deque as (bs' ++ reverse as' ++ bs)
```

io-sim/src/Data/Deque/Strict.hs
```diff
@@ -0,0 +1,41 @@
...
+
+instance Monoid (Deque a) where
+    mempty = Deque [] []
+
+instance Foldable Deque where
+    foldr step init (Deque head tail) =
+      foldr step (foldr step init (reverse tail)) head
```
bolt12, line 22: We can avoid reversing here with `foldl' (flip step)`

io-sim/src/Data/Deque/Strict.hs
```diff
@@ -0,0 +1,41 @@
...
+    foldl' step init (Deque head tail) =
+      foldl' step (foldl' step init head) (reverse tail)
```
bolt12, line 24: Same here with `foldr' (flip step)`

bolt12 (changes requested):
I think I found some things that need fixing

io-sim/src/Data/Deque/Strict.hs
coot: You're right, I messed this up after trying to do that on the head side.

io-sim/src/Data/Deque/Strict.hs
```diff
@@ -0,0 +1,41 @@
...
+      Deque as (bs' ++ reverse as' ++ bs)
+
+instance Monoid (Deque a) where
+    mempty = Deque [] []
+
+instance Foldable Deque where
+    foldr step init (Deque head tail) =
+      foldr step (foldl' (flip step) init tail) head
+    foldl' step init (Deque head tail) =
+      foldr (flip step) (foldl' step init head) tail
```
bolt12, line 24: I think this should be `foldr'` since we are implementing `foldl'`

bolt12 (approved):
Just one small thing, I'll approve in the meantime :)

bolt12 (approved):

Issue #76, opened 2023-03-22
IOSimPOR assertion failures


coot:
Recently two execution of the `propExploration` errored:
* https://github.com/input-output-hk/io-sim/actions/runs/4491097750/jobs/7899072815
* https://github.com/input-output-hk/io-sim/actions/runs/4413935534/jobs/7735014212?pr=73

## Logs

```
    propExploration:                           FAIL (36.64s)
      *** Failed! (after 53 tests and 48 shrinks):
      Exception:
        Assertion failed
        CallStack (from HasCallStack):
          assert, called at src/Control/Monad/IOSimPOR/Internal.hs:990:5 in io-sim-0.4.0.0-inplace:Control.Monad.IOSimPOR.Internal
      Tasks [Task [WhenSet 9 0],Task [WhenSet 0 0],Task [WhenSet 10 10,WhenSet 10 9],Task [WhenSet 0 0]]
      Schedule control: ControlAwait [ScheduleMod (RacyThreadId [4],1) ControlDefault [(RacyThreadId [3],0),(RacyThreadId [3],1),(RacyThreadId [2],0),(RacyThreadId [2],1),(RacyThreadId [1],0),(RacyThreadId [1],1),(RacyThreadId [2],2)]]
      RacyThreadId [3] delayed at time Time 0s
        until after:
          RacyThreadId [4]
      
[...]

PR #79, opened 2023-04-04
Fix issue #76

bolt12:
Ignore effectStatusWrites in order to not generate bad schedules that lead to assertion failures.

Also fix issue introduced in 669d6b23c41d22f503916b7a924a11c39b442347 where the onlyReadEffect is not checking for effectStatusRead effects.

Check #76 for more details

### DIFF
--- a/io-sim/src/Control/Monad/IOSimPOR/Types.hs
+++ b/io-sim/src/Control/Monad/IOSimPOR/Types.hs
@@ -64,8 +64,17 @@ statusWriteEffects tids = mempty{effectStatusWrites = tids}
 someTvarId :: SomeTVar s -> TVarId
 someTvarId (SomeTVar r) = tvarId r
 
+-- | Make sure we only have read effects
+--
+-- We ignore `effectStatusReads` and `effectStatusWrites`. This is because
+-- effects of `threadStatus` are non-blocking, if they block it is a result of
+-- of a blocking stm transaction.
+--
 onlyReadEffect :: Effect -> Bool
-onlyReadEffect e = e { effectReads = effectReads mempty } == mempty
+onlyReadEffect e = e { effectReads        = effectReads mempty
+                     , effectStatusReads  = effectStatusReads mempty
+                     , effectStatusWrites = effectStatusWrites mempty
+                     } == mempty
 
 racingEffects :: Effect -> Effect -> Bool
 racingEffects e e' =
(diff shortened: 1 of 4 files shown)


io-sim/src/Control/Monad/IOSimPOR/Types.hs
```diff
@@ -64,8 +64,17 @@ statusWriteEffects tids = mempty{effectStatusWrites = tids}
 someTvarId :: SomeTVar s -> TVarId
 someTvarId (SomeTVar r) = tvarId r
 
+-- | Make sure we only have read effects
+--
+-- It seems we need to ignore statusWrites effects to avoid generating bad
+-- schedules. See issue #76 for examples.
```
coot, line 70: We must ignore the `statusWrite` effects.   This function is used to only limit the blocking steps which are blocked on `STM`.

Maybe this function should be written as:
```
onlyReadEffect e = effectWrite e == mempty
```
but I am not 100% sure.

coot: If you update the comment we can merge it.

coot (approved):
LGTM

io-sim/src/Control/Monad/IOSimPOR/Types.hs
coot: @bolt12 will the `onlyReadEffect` that I suggest above fix the test case which you mentioned here: https://github.com/input-output-hk/io-sim/issues/76#issuecomment-1495946252

bolt12: The fix I have fixes the test case mentioned in #76 . I am going to try your suggestion

io-sim/src/Control/Monad/IOSimPOR/Types.hs
```diff
@@ -82,9 +82,11 @@ racingEffects e e' =
...
    || effectWrites e       `intersects`  effectReads  e'
    || effectWrites e       `intersects`  effectWrites e'
-   || effectStatusReads e  `intersectsL` effectStatusWrites e'
-   || effectStatusWrites e `intersectsL` effectStatusReads  e'
-   || effectStatusWrites e `intersectsL` effectStatusWrites e'
+   {--
+     - || effectStatusReads e  `intersectsL` effectStatusWrites e'
+     - || effectStatusWrites e `intersectsL` effectStatusReads  e'
+     - || effectStatusWrites e `intersectsL` effectStatusWrites e'
+     --}
```
bolt12, line 89: Should we leave a TODO and corresponding issue to get back to this?

bolt12 (commented):
LGTM!

coot:
Closing in favour of removal of `threadStatus`.

PR #82, opened 2023-04-13
timeouts and exceptions

coot:
- io-sim: fixed timeouts in presence of exceptions
- io-sim: code style
- io-sim: track reason why a thread is blocked
- io-sim: refactored timeout implementation
- io-sim: improved TimeoutException Show instance

### DIFF
--- a/io-sim/src/Control/Monad/IOSim/Internal.hs
+++ b/io-sim/src/Control/Monad/IOSim/Internal.hs
@@ -1000,10 +1000,10 @@ unwindControlStack :: forall s a.
                    -> ( Either Bool (Thread s a)
                       , Timeouts s
                       )
-unwindControlStack e thread timers0 =
+unwindControlStack e thread = \timers ->
     case threadControl thread of
       ThreadControl _ ctl ->
-        unwind (threadMasking thread) ctl timers0
+        unwind (threadMasking thread) ctl timers
   where
     unwind :: forall s' c. MaskingState
            -> ControlStack s' c a
--- a/io-sim/src/Control/Monad/IOSim/Internal.hs
+++ b/io-sim/src/Control/Monad/IOSim/Internal.hs
@@ -128,6 +128,9 @@ data TimerCompletionInfo s =
      -- ^ `timeout` timer run by `ThreadId` which was assigned the given
      -- `TimeoutId` (only used to report in a trace).
 
+
+type Timeouts s = OrdPSQ TimeoutId Time (TimerCompletionInfo s)
+
 -- | Internal state.
 --
 data SimState s a = SimState {
@@ -140,7 +143,7 @@ data SimState s a = SimState {
        -- | current time
        curTime  :: !Time,
        -- | ordered list of timers and timeouts
-       timers   :: !(OrdPSQ TimeoutId Time (TimerCompletionInfo s)),
+       timers   :: !(Timeouts s),
        -- | list of clocks
        clocks   :: !(Map ClockId UTCTime),
        nextVid  :: !TVarId,     -- ^ next unused 'TVarId'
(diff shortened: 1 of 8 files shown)


coot:
The following tests have passed:
* [ ] `network-mux:test`
* [ ] `ouroboros-network-framework:test`
* [ ] `ouroboros-network:test` (including nightly ones)

io-sim/src/Control/Monad/IOSim/Internal.hs
```diff
@@ -129,7 +131,7 @@ data TimerCompletionInfo s =
      | TimerThreadDelay !ThreadId !TimeoutId
      -- ^ `threadDelay` timer run by `ThreadId` which was assigned the given
      -- `TimeoutId` (only used to report in a trace).
-     | TimerTimeout !ThreadId !TimeoutId !(STRef s IsLocked)
+     | TimerTimeout !ThreadId !TimeoutId !(TMVar (IOSim s) ThreadId)
```
bolt12, line 134: Why changing to a TMVar?

io-sim/src/Control/Monad/IOSim/Internal.hs
```diff
@@ -244,47 +246,27 @@ schedule !thread@Thread{
...
+            threadAction :: IOSim s ()
+            threadAction =
+              if v then unsafeUnregisterTimeout tmid
+                   else atomically (takeTMVar lock) >>= killThread
+
+            thread' =
+              thread { threadControl =
+                        ThreadControl (case threadAction of
+                                        IOSim k' -> k' (\() -> k (Just x)))
+                                      ctl'
```
bolt12, line 267: Why do you need pattern matching here?

io-sim/src/Control/Monad/IOSimPOR/Internal.hs
```diff
@@ -1623,6 +1598,15 @@ execWriteTVar :: TVar s a -> a -> ST s ()
 execWriteTVar TVar{tvarCurrent} = writeSTRef tvarCurrent
 {-# INLINE execWriteTVar #-}
 
+execTryPutTMVar :: TVar s (Maybe a) -> a -> ST s Bool
+execTryPutTMVar var a = do
```
bolt12, line 1602: This function is called `execTryPutTMVar` but receives a `TVar`

bolt12 (approved):

io-sim/src/Control/Monad/IOSim/Internal.hs
coot: We need `SimA` not `IOSim`.

io-sim/src/Control/Monad/IOSim/Internal.hs
coot: The primary reason was that I wanted to write thread actions using `IOSim` not `SimA`.  It turned out that I had to use the `execReadTVar` but in a limited way.

We actually should use an `MVar`, but it doesn't matter in our use case.  We have only one thread that runs `takeMVar`, so we don't need fairness.  We have two threads that try to write to it, but it's guaranteed that one of them receives the `TMVar` empty, so fairness is not a problem here either.

io-sim/src/Control/Monad/IOSimPOR/Internal.hs
coot: The alternative is:
```
execTryPutTMVar :: TMVar (IOSim s) a -> a -> ST s Bool
```

io-sim/src/Control/Monad/IOSim/Internal.hs
bolt12: Okay got it! I don't mind the change!

io-sim/src/Control/Monad/IOSim/Internal.hs
bolt12: What do you mean with that? What about the other places where we don't do pattern match? It is not clear to me why in this case we need to do this

io-sim/src/Control/Monad/IOSimPOR/Internal.hs
coot: I changed to `TMVar`s.

io-sim/src/Control/Monad/IOSim/Internal.hs
coot: because in other palces we use `SimA` to write thee syntax tree.

bolt12: Oh nevermind forgot that threadAction existed and it's of type IOSim.

PR #86, opened 2023-04-14
threadDelay: efficiently remove thread delay timer

coot:
Unfortunately we cannot track the information on which `ThreadDelay`
a thread is blocked in `threadStatus` because it is cleared when
a thread is rescheduled.  This is too early.  In the scenario we are
interested in, the thread is rescheduled when the thread receives an
exception and it's when we interpret the `Throw` we need to know the
`TimeoutId` of that `threadDelay`.

We add a `threadDelayId` field, which tracks the last `threadDelay` id.
This field only cleared when the thread receives an exception, thus it's
life time is long enough to be useful.

### DIFF
--- a/io-sim/src/Control/Monad/IOSim/Internal.hs
+++ b/io-sim/src/Control/Monad/IOSim/Internal.hs
@@ -263,8 +263,13 @@ schedule !thread@Thread{
                      }
         schedule thread' simstate
 
+      DelayFrame tmid k ctl' -> do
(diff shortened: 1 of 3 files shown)


bolt12 (approved):

bolt12 (approved):

dcoutts (commented):
Noice!

coot:
I hope the noise isn't too laud :grin:.

PR #88, opened 2023-04-18
Move `Control.Monad.Class.MonadMVar` to `Control.Concurrent.Class.MonadMVar`

jorisdral:
As suggested in https://github.com/input-output-hk/io-sim/pull/74#discussion_r1156071137, we move `Control.Monad.Class.MonadMVar` to `Control.Concurrent.Class.MonadMVar`, which is closer to what `base` uses in its module hierarchy (`Control.Concurrent.MVar`), and where `MonadSTM` lives (`Control.Concurrent.Class.MonadSTM`).

Question: should we leave deprecation warnings behind in the old modules? What is the general policy for deprecation in this repo?

### DIFF
--- a/io-classes/src/Control/Concurrent/Class/MonadMVar.hs
+++ b/io-classes/src/Control/Concurrent/Class/MonadMVar.hs
@@ -3,7 +3,7 @@
 {-# LANGUAGE TypeFamilyDependencies #-}
 {-# LANGUAGE TypeOperators          #-}
 
-module Control.Monad.Class.MonadMVar (MonadMVar (..)) where
+module Control.Concurrent.Class.MonadMVar (MonadMVar (..)) where
 
 import qualified Control.Concurrent.MVar as IO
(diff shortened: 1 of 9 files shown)


io-classes/src/Control/Concurrent/Class/MonadMVar.hs
```diff
@@ -3,7 +3,7 @@
 {-# LANGUAGE TypeFamilyDependencies #-}
 {-# LANGUAGE TypeOperators          #-}
 
-module Control.Monad.Class.MonadMVar (MonadMVar (..)) where
+module Control.Concurrent.Class.MonadMVar (MonadMVar (..)) where
```
coot, line 6: Could you add the `Control.Monad.Class.MonadMVar` and reexport from it the new one but add a deprecation warning to it.

coot (commented):
This looks good.   Could you add a changelog entry.

coot (approved):
LGTM

PR #92, opened 2023-04-26
coot/ghc 8.10 fix

coot:
- io-classes: fixed building haddocks with ghc-8.10
- version 1.1.0.0

### DIFF
--- a/io-classes-mtl/io-classes-mtl.cabal
+++ b/io-classes-mtl/io-classes-mtl.cabal
@@ -46,7 +46,7 @@ library
                       array,
                       mtl,
 
-                      io-classes ^>= 1.0.0.0,
+                      io-classes  >= 1.0 && < 1.2,
                       si-timers,
 
 
--- a/io-classes/CHANGELOG.md
+++ b/io-classes/CHANGELOG.md
@@ -1,11 +1,15 @@
 # Revsion history of io-classes
 
-## Next version
+## 1.1.0.0
 
 ### Breaking changes
 
 * `Control.Monad.Class.MonadMVar` is now deprecated in favour of
   `Control.Concurrent.Class.MonadMVar`.
+
+### Non breaking changes
+
+* Fixed building haddocks with `ghc-8.10`.
 
 ## 1.0.0.1
 
--- a/io-sim/io-sim.cabal
+++ b/io-sim/io-sim.cabal
@@ -1,6 +1,6 @@
 cabal-version:       3.0
 name:                io-sim
-version:             1.0.0.1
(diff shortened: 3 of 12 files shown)


io-classes/CHANGELOG.md
jorisdral, line 1: `## Next version` should be replaced by `## 1.1.0.0`

io-classes-mtl/io-classes-mtl.cabal
```diff
@@ -46,7 +46,7 @@ library
                       array,
                       mtl,
 
-                      io-classes ^>= 1.0.0.0,
+                      io-classes  >= 1.0 && < 1.2,
```
jorisdral, line 49: There is no changelog entry for this change

si-timers/si-timers.cabal
```diff
@@ -59,7 +59,7 @@ library
                        stm,
                        time              >=1.9.1 && <1.13,
 
-                       io-classes       ^>=1.0
+                       io-classes        >=1.0 && <1.2
```
jorisdral, line 62: If this is the only change made to the package between `1.0.0.0` and `1.1.0.0`, would a revision of `1.0.0.0` be sufficient? Or is it the goal to release the packages in lockstep with the same version for each package?

io-sim/io-sim.cabal
```diff
@@ -77,14 +77,14 @@ library
                        ScopedTypeVariables,
                        TypeFamilies
   build-depends:       base              >=4.9 && <4.19,
-                       io-classes       ^>=1.0,
+                       io-classes       ^>=1.1,
```
jorisdral, line 80: Would `>= 1.0 && < 1.2` also suffice? I think building with `1.0` would cause some deprecation warnings to be emitted with respect to moving the `MonadMVar` module, but would it lead to actual build failures?

io-classes/CHANGELOG.md
coot: Fixed.

io-classes-mtl/io-classes-mtl.cabal
coot: Yes, because I will just add a revision on hackage, no need to bump its version.

si-timers/si-timers.cabal
coot: I do release `io-classes`, `io-sim`, `strict-stm`, `strict-mvar` and `si-timers` (but not `io-classes-mtl`) in lock step, but I am fine to be more relaxed with lower bounds when I am sure things will still work.

PR #100, opened 2023-05-18
strict-stm: Fix typo'd checktvarinvariant CPP flag

fraser-iohk:
Fixes the CPP flag being misnamed, causing the invariant checking code to never build


jasagredo (approved):
A well-hidden typo 😤

PR #102, opened 2023-05-29
Test that values inside `StrictMVar`s are in WHNF

jorisdral:


jorisdral:
@coot what do you think of the setup of the tests in this PR? I'm mainly uncertain about whether to duplicate the tests for unchecked and checked `StrictMVar`s, or to use a `cabal` flag

jorisdral:
This PR should add a changelog entry

strict-mvar/test/Test/Control/Concurrent/Class/MonadMVar/Strict/WHNF.hs
```diff
@@ -0,0 +1,201 @@
...
+infixr 9 .:
+
+(.:) :: (y -> z) -> (x0 -> x1 -> y) -> (x0 -> x1 -> z)
+(.:) g f x0 x1 = g (f x0 x1)
+
+isInWHNF :: (MonadMVar m, Typeable a) => StrictMVar m a -> PropertyM m Bool
+isInWHNF v = do
+    x <- run $ readMVar v
+    let tinfoMay = unsafeNoThunks (OnlyCheckWhnf x)
+    case tinfoMay of
```
coot, line 94:
```suggestion
    case unsafeNoThunks (OnlyCheckWhnf x) of
```

strict-mvar/test/Test/Control/Concurrent/Class/MonadMVar/Strict/WHNF.hs
```diff
@@ -0,0 +1,201 @@
...
+#else
+name = "Strict"
+#endif
+
+tests :: TestTree
+tests = testGroup ("Test.Control.Concurrent.Class.MonadMVar." <> name) [
+      testGroup "WHNF" [
+          testGroup "IO" [
+              testProperty "prop_newMVar" $
+                monadicIO .: prop_newMVar (Proxy @Int)
```
coot, line 40: Is it worth to keep the tests, e.g. `prop_newMVar` polymorphic if we only instantiate them with `Int`?  I think it would be cleaner to monomorphise them.

coot (approved):

Issue #103, opened 2023-05-29
`strict-mvar`: propagate `HasCallStack` constraints


jorisdral:
In `Control.Concurrent.Class.MonadMVar.Strict.Checked`, we check invariants using:

```haskell
checkInvariant :: HasCallStack => Maybe String -> a -> a
```

However, none of the functions that use this have `HasCallStack` constraints themselves. We should add these. We should also add them to `Control.Concurrent.Class.MonadMVar.Strict`.

PR #105, opened 2023-05-31
Propage `HasCallStack` constraint when using `checkInvariant`.

jorisdral:
See #103.


coot (approved):
I am fine, although I don't like `HasCallStack` myself.  Let's see if @dcoutts agrees or not :grin:.

PR #106, opened 2023-05-31
New `MonadInspectMVar` class

jorisdral:
Add a new `MonadInspectMVar` class (similar to `MonadInspectSTM`) with an `inspectMVar` function for accessing an `MVar` in an underlying monad (if applicable). This is mainly useful for `io-sim`, since the underlying monad is `ST`. `IO` has no underlying monad, so the provided instance for `IO` defaults `inspectMVar` to `tryReadMVar`.

### DIFF
--- a/io-classes/src/Control/Concurrent/Class/MonadMVar.hs
+++ b/io-classes/src/Control/Concurrent/Class/MonadMVar.hs
@@ -1,9 +1,12 @@
 {-# LANGUAGE DefaultSignatures      #-}
+{-# LANGUAGE FlexibleContexts       #-}
 {-# LANGUAGE QuantifiedConstraints  #-}
 {-# LANGUAGE TypeFamilyDependencies #-}
-{-# LANGUAGE TypeOperators          #-}
 
-module Control.Concurrent.Class.MonadMVar (MonadMVar (..)) where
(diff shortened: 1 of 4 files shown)


io-classes/src/Control/Concurrent/Class/MonadMVar.hs
```diff
@@ -181,8 +187,22 @@ instance ( MonadMask m
     modifyMVarMasked (WrappedMVar v) f = ReaderT $ \r ->
       modifyMVarMasked v (\a -> runReaderT (f a) r)
 
+--
+-- MonadInspectMVar
+--
 
-
+-- | This type class is intended for
+-- ['io-sim'](https://hackage.haskell.org/package/io-sim), where one might want
```
coot, line 195: oh, does haddock accept markdown style links?

coot: It does :)

coot (approved):
LGTM

io-classes/src/Control/Concurrent/Class/MonadMVar.hs
jorisdral: Yeah, I copied this from `MonadInspectSTM`!

PR #108, opened 2023-06-14
Factor out checked `StrictMVar`s

jorisdral:
This PR removes checked `StrictMVar`s from `strict-mvar` by reverting #74, #105 and parts of  #102. Checked strict-mvars will live in [the `ouroboros-consensus` repository](https://github.com/input-output-hk/ouroboros-consensus/) instead.


coot (commented):
Please update the changelog file.

jorisdral:
> Please update the changelog file.

I did that I think, what do you think about its contents?

jorisdral:
> LGTM, would you also remove the checks from strict-stm?

Yes, but I'll probably open a separate PR for this

PR #115, opened 2023-09-28
Refactored package structure: use sublibraries

coot:
Put `si-timers`, `strict-stm` and `strict-mvar` as public libraries of `io-classes`.

`strict-mvar:test` is left in `test-strict-mvar` as it introduces
a cycle between `io-sim` and `io-classes` which cabal cannot resolve.


bolt12 (approved):
LGTM, needs a changelog entry

coot:
It would be nice to get `cabal` changes to the `haddock-project` command before we merge it.  I looked into and it seems doable, although requires a bit of thinking to get all the parts right.  Once new version of `cabal` is published we can use it to publish haddocks on Github Pages, until Hackage also supports public sub-libraries (which if I am lucky :wink: I will also fix in the same set of changes).

jorisdral:
> strict-mvar:test is left in test-strict-mvar as it introduces a cycle between io-sim and io-classes which cabal cannot resolve.

One alternative would be to split the tests, moving the ones for IO to `io-classes`, and the ones for `IOSim` to `io-sim`

coot:
@jorisdral done;

`cabal` PR haskell/cabal#9821 was merged, so with next `cabal` release we can merge it.

PR #120, opened 2023-11-08
Caching on GHA

coot:
We are downloading `stylish-haskell` rather than building it, so no need
to use `actions/cache`.


coot:
The commit 07d1016 seems to fix actions/cache#1275, which is a bit strange.

bolt12 (approved):

PR #131, opened 2024-01-02
Mostly haddocks, but also Alternative instance for IOSim

coot:
* improved haddocks
* removed deprecated modules
* added `Alternative` and `MonadPlus` instances for `IOSim`


bolt12 (approved):
LGTM

PR #135, opened 2024-01-30
Remove wrong implementation of TQueue in io-sim

### LINKED ISSUE #133: flushTQueue is broken
**Describe the bug**
`flushTQueue` in `IOSim s` does not empty the queue, while it does in `IO`

**To Reproduce**

Execute this program:

```haskell
#!/usr/bin/env cabal
{- cabal:
build-depends: base, io-sim ^>= 1.3, io-classes ^>= 1.3
-}

import Control.Concurrent.Class.MonadSTM
import Control.Exception (assert)
import Control.Monad (unless)
import Control.Monad.IOSim

main :: IO ()
main = do
  emptyInIO <- emptyQueueAfterFlush
  unless emptyInIO $
    error "queue not emptied from IO"

  let emptyInIOSim = runSimOrThrow emptyQueueAfterFlush
  unless emptyInIOSim $
    error "queue not emptied from IOSim"

emptyQueueAfterFlush :: MonadSTM m => m Bool
emptyQueueAfterFlush = do
  q <- newTQueueIO
  atomically $ do
    writeTQueue q 1
    _ <- flushTQueue q
    isEmptyTQueue q
```

**Expected behaviour**
Same semantics of `STM` in `IO` and `IOSim`

**Additional context**
The [`flushQueueDefault`](https://hackage.haskell.org/package/io-sim-1.3.1.0/docs/src/Control.Monad.IOSim.STM.html#flushTQueueDefault) looks wrong when compared with the one of [io-classes](https://hackage.haskell.org/package/io-classes-1.3.1.0/docs/src/Control.Monad.Class.MonadSTM.Internal.html#flushTQueueDefault) and the real one from [stm](https://hackage.haskell.org/package/stm-2.5.3.0/docs/src/Control.Concurrent.STM.TQueue.html#flushTQueue)

ch1bo:
Fixes #133 by using the default implementation of MonadSTM.

TBD: remove the TBQueue implementation as well?


io-classes/src/Control/Monad/Class/MonadSTM/Internal.hs
```diff
@@ -827,7 +821,24 @@ flushTQueueDefault (TQueue read write) = do
...
+  :: MonadTraceSTM m
+  => proxy m
+  -> TQueueDefault m a
+  -> (Maybe [a] -> [a] -> InspectMonad m TraceValue)
+  -> STM m ()
+traceTQueueDefault p (TQueue read write) f = do
+    traceTVar p read
+              (\mas as -> f mas as)
+    traceTVar p write
+              (\mas as -> f mas as)
```
coot, line 841: This is wrong, because `f` has access to either side (read / write) of the queue, rather than all its elements.  This is the reason why I kept the default implementation of `TQueueDefault` in a single `TVar` instead of two.  This approach requires more substantial changes to `io-sim`.

I think it's much simpler to just fix `flushTQueueDefault`.

coot (changes requested):

io-classes/src/Control/Monad/Class/MonadSTM/Internal.hs
ch1bo: So you suggest keeping the differing implementation using a single `TVar`? Why is tracing a deciding factor here - can't we implement this on the standard implementation of read/write end queues?

coot: Because it is designed for verification purposes, so it's quite crucial for the library.

> So you suggest keeping the differing implementation using a single TVar?

Yes, keep the current implementation which btw is based on a standard queue data type (see Okasaki's book).  `stm` splits it into two `TVar`'s I think for optimisation purposes, e.g. some operations will only need access one of the `TVar`s which might change the contention in a parallel scenario - in `io-sim` we don't have any parallelism, so it's not a factor.

coot: One invariant that one might want to be able to verify is the number of elements in a queue (if the application has such an invariant).  It will be impossible to write it with the proposed `traceTQueueDefault`.

ch1bo: Okay, let me try to write a `flushTQueue` which works on the single `TVar`.

I don't understand `traceTQueueDefault` and if you say we cannot implement it correctly for the standard double TVar implementation, I need to trust you. 

But I do see the continued risk of maintaining a "differing" implementation on `io-sim` which does not correspond to the IO/STM variant semantics - if we could "just" keep lifted copies of the `base` and `stm` implementations around, that would be much more maintainable by casual contributors like me.

ch1bo: Addressed in https://github.com/ch1bo/io-sim/commit/9e2444d3e22e74ce9eea3fb7c0dc7799d63e3ec3, but while doing so I discovered another bug (which underlines my worry of having non-standard implementation): https://github.com/input-output-hk/io-sim/pull/136

abailly-iohk: Would it not make sense to have an "equivalence property" written for all `io-sim` functions that "guarantees" the semantics is consistent with what's provided by the `IO` instance?

ch1bo:
All comments addressed on this one @coot

io-classes/src/Control/Monad/Class/MonadSTM/Internal.hs
coot: @abailly-iohk that's a good idea.

io-sim/src/Control/Monad/IOSim/STM.hs
```diff
@@ -95,7 +95,10 @@ tryPeekTQueueDefault (TQueue queue) = do
       []   -> Nothing
 
 flushTQueueDefault :: MonadSTM m => TQueueDefault m a -> STM m [a]
-flushTQueueDefault (TQueue queue) = uncurry (++) <$> readTVar queue
+flushTQueueDefault (TQueue queue) = do
+  xs <- uncurry (++) <$> readTVar queue
```
coot, line 99:
```suggestion
  xs <- (\head tail -> head ++ reverse tail) <$> readTVar queue
```

coot (changes requested):

io-classes/src/Control/Monad/Class/MonadSTM/Internal.hs
coot: I created an issue for this: io-sim/issues#137.

coot (approved):
LGTM

io-classes/src/Control/Monad/Class/MonadSTM/Internal.hs
coot: > I don't understand traceTQueueDefault and if you say we cannot implement it correctly for the standard double TVar implementation, I need to trust you.

Just for the sake of explanation:

`traceTVar` guarantees that its callback is called when the `TVar` is committed in an stm transaction and receives the current (and previous) data stored in the `TVar`.  If `TQueueDefault` is using two `TVar`s `io-sim` will need to know that not only the committed `TVar` matters but also the other one.  That's the reason why two `TVar`s are problematic.

`traceTVar` is designed for writing properties when order of events matters but one cannot relay on order of trace messages (i.e. in concurrent scenario order of trace messages will depend on the scheduler). `traceTVar` callbacks are guaranteed to be called right after an atomic transaction is committed in this since they do not depend on the scheduler. If you do `readTVarIO >>= traceM . show`, `io-sim` will deschedule the thread after `readTVarIO` and you're no longer guaranteed that multiple concurrent `traceM` calls will be done in the same order as the atomic transactions.

PR #136, opened 2024-01-30
WIP io-sim: flushTQueue does not maintain order

ch1bo:
Quick reproduction to confirm another bug in `flushTQueue` discovered while working on #133


io-sim/io-sim.cabal
```diff
@@ -1,6 +1,6 @@
 cabal-version:       3.0
 name:                io-sim
-version:             1.3.1.0
+version:             1.3.1.1
```
coot, line 3: Could you revert this, I figure out version changes when releasing to hackage.

coot (changes requested):
LGTM, just a minor comment.

io-sim/src/Control/Monad/IOSim/STM.hs
```diff
@@ -95,7 +95,10 @@ tryPeekTQueueDefault (TQueue queue) = do
       []   -> Nothing
 
 flushTQueueDefault :: MonadSTM m => TQueueDefault m a -> STM m [a]
-flushTQueueDefault (TQueue queue) = uncurry (++) <$> readTVar queue
+flushTQueueDefault (TQueue queue) = do
+  xs <- uncurry (++) <$> readTVar queue
```
coot, line 99: I think you need to reverse the tail, e.g. 
```suggestion
  xs <- (\front tail -> front ++ reverse tail) <$> readTVar queue
```

ch1bo: I know.. I was not trying to fix it yet, but get the other PR merged first.

PR #140, opened 2024-02-12
QuickCheck monadic combinators

coot:


io-sim/src/Control/Monad/IOSim.hs
```diff
@@ -751,3 +758,73 @@ compareTracesST (Just passing) trace = do
...
+--   nat :: forall s a. State -> M s a -> 'IOSim' s a
+--   nat state m = runStateT (runM m) state
+-- @
+--
+-- @since 1.4.1.0
+--
+monadicIOSim :: (Testable a, forall s. Monad (m s))
+             => (SimTrace Property -> Property)
+             -- ^ Allows to trace `SimTrace` in counterexamples.  The simplest
+             -- use case is to pass which `monadicIOSim` does.
```
bolt12, line 809:
```suggestion
             -- use case is to pass which `monadicIOSim_` does.
```

bolt12 (approved):

io-sim/src/Control/Monad/IOSim.hs
coot: Any better idea for the two names `monadicIOSim_` and `monadicIOSim`.  I put them this way because `monadicIOSim` might be actually more useful than the specialised one since it's useful when dealing with monad transformer stacks.

PR #143, opened 2024-02-14
Prevent STM waking up threads blocked on `threadDelay`

### LINKED ISSUE #142: a registerDelay timeout is truncating a threadDelay call
**Describe the bug**
A `registerTimeout` that happens to be ignored by the thread ends up truncating a subsequent `threadDelay` in the same thread.

**To Reproduce**
Steps to reproduce the behaviour:

My thread is calling `registerTimeout`, calling `atomically` on a transaction that is blocked on the timeout's `TVar Bool` and others, completing based on changes to the others, and then later calling `threadDelay maxBound`. When the `registerTimeout` fires, the thread is woken up (it happens to be after 2s, in this case) --- despite `maxBound` not having elapsed (there are other threads in my `runIOSim*` invocation).

I have a large repro, but there are lots of dependencies. I do not anticipate being able to easily minimize it. It involves a `SI.registerDelay 10` and `threadDelay maxBound` (happens with or without the `si-timers` interface).

(I suspect the minimal repro needs to involve an STM transaction that would be retried when the timeout fires but ends up committing before that happens (because some other tvar changes in such a way that the makes the timeout irrelevant.)

**Expected behaviour**
Both:

- Nothing should truncate a call `threadDelay`.
- Once the timeout's `TVar` becomes dead code, nothing should be affected by the timeout triggering. (In my example, I would have happily called, say, `clearTimeout` if that kind of feature were already part of the `io-classes` API. Which actually would have masked this bug, I suspect!)

**Desktop (please complete the following information):**
```
$ cat cabal.project.freeze  | grep -e io-sim -e io-classes -e ghc | grep -e ==
             any.ghc-bignum ==1.2,
             any.ghc-boot-th ==9.2.8,
             any.ghc-heap ==9.2.8,
             any.ghc-prim ==0.8.0,
             any.io-classes ==1.3.1.0,
             any.io-classes-mtl ==0.1.0.2,
             any.io-sim ==1.3.1.0,
```
**Additional context**
Add any other context about the problem here. Attach `io-sim` or `io-sim-por` trace if possible.

I'll attach the trace as the first GitHub comment.

amesgen:
Closes #142

### DIFF
--- a/io-sim/CHANGELOG.md
+++ b/io-sim/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## next version (1.4.1.0)
 
+* Prevent STM waking up threads blocked on `threadDelay`.
+
 ### Non-breaking changes
 
 * QuickCheck monadic combinators: `monadicIOSim`, `monadicIOSim_` and `runIOSimGen`.
--- a/io-sim/src/Control/Monad/IOSim/Internal.hs
+++ b/io-sim/src/Control/Monad/IOSim/Internal.hs
@@ -747,8 +747,10 @@ reschedule !simstate@SimState{ threads, timers, curTime = time } =
 
             -- Check all fired threadDelays
         let wakeupThreadDelay = [ (tid, tmid) | TimerThreadDelay tid tmid <- fired ]
-            wakeup            = fst `fmap` wakeupThreadDelay ++ wakeupSTM
-            (_, !simstate')   = unblockThreads False wakeup simstate
+            !simstate'        =
+                snd . unblockThreads False (fst `fmap` wakeupThreadDelay)
+              . snd . unblockThreads True  wakeupSTM
+              $ simstate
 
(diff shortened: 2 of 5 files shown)


bolt12 (approved):

coot (commented):
Thanks @amesgen, that's a clean fix.

This is actually half of the work :grin:, would you like to do the `IOSimPOR` part too?

nfrisby:
This fix works for me! My new GSM tests are passing locally. My PR is unblocked (other than needing a CHaP release, etc---I've merely locally backported Esgen's patch to the last `io-sim` release and used a `source-repository-package` in my local `cabal.project` for now).

coot:
I'll cut a new `hackage` release, once #141 is merged.

coot (commented):
LGTM, thanks @amesgen :tada:

coot (approved):

PR #147, opened 2024-02-23
io-classes: MonadST a subclass of PrimMonad

coot:


nfrisby (approved):
Thanks, this diff does avoid the confusion that I was referring to in the Issue I opened.

Issue #151, opened 2024-05-02
Add more functions to Data.List.trace


coot:
It would be nice to have things like:

* `take :: Int -> Trace a b -> Trace a ()`
* `takeWhile :: (a -> Bool) -> Trace a b -> Trace a ()`
* `drop :: Int -> Trace a b -> Trace a b`
* `dropWhile :: (a -> Bool) -> Trace a b -> Trace a b`

PR #153, opened 2024-05-03
Revert commit 4534b6eae64072a87bd81584f479a123681358a3

bolt12:
This introduced a regression where IOSimPOR could not handle infinite simulations.

Closes: #151

### DIFF
--- a/io-sim/src/Control/Monad/IOSim.hs
+++ b/io-sim/src/Control/Monad/IOSim.hs
@@ -296,8 +296,8 @@ printTraceEventsSay = mapM_ print . selectTraceEventsSay
 --
 traceSelectTraceEvents
     :: (Time -> SimEventType -> Maybe b)
-    -> SimTrace a
-    -> Trace (SimResult a) b
+    -> Trace a SimEvent
+    -> Trace a b
 traceSelectTraceEvents fn = bifoldr ( \ v _acc -> Nil v )
                                     ( \ eventCtx acc
                                      -> case eventCtx of
@@ -316,7 +316,8 @@ traceSelectTraceEvents fn = bifoldr ( \ v _acc -> Nil v )
 -- | Select dynamic events.  It is a /total function/.
 --
 traceSelectTraceEventsDynamic :: forall a b. Typeable b
-                              => SimTrace a -> Trace (SimResult a) b
+                              => Trace a SimEvent
+                              -> Trace a b
 traceSelectTraceEventsDynamic = traceSelectTraceEvents fn
   where
     fn :: Time -> SimEventType -> Maybe b
@@ -326,7 +327,7 @@ traceSelectTraceEventsDynamic = traceSelectTraceEvents fn
 
 -- | Select say events.  It is a /total function/.
 --
-traceSelectTraceEventsSay :: forall a.  SimTrace a -> Trace (SimResult a) String
+traceSelectTraceEventsSay :: forall a. Trace a SimEvent -> Trace a String
 traceSelectTraceEventsSay = traceSelectTraceEvents fn
   where
     fn :: Time -> SimEventType -> Maybe String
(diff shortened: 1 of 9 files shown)


io-sim/src/Control/Monad/IOSim.hs
```diff
@@ -619,42 +574,59 @@ exploreSimTraceST optsf main k =
...
 
 
 -- |  Trace `SimTrace` to `stderr`.
 --
-traceDebugLog :: Int -> SimTrace a -> ST s ()
-traceDebugLog logLevel _trace | logLevel <= 0 = pure ()
-traceDebugLog 1 trace = Debug.traceM $ "Simulation trace with discovered schedules:\n"
-                                    ++ Trace.ppTrace (ppSimResult 0 0 0) (ppSimEvent 0 0 0) (ignoreRaces $ void `first` trace)
-traceDebugLog _ trace = Debug.traceM $ "Simulation trace with discovered schedules:\n"
-                                    ++ Trace.ppTrace (ppSimResult 0 0 0) (ppSimEvent 0 0 0) (void `first` trace)
```
coot, line 656: It would be nice to keep `traceDebugLog` in some form, it was used in `exploreSimTrace`.  Note that it's implemented with `Debug.traceM` so it can run in any monad, if we use `Debug.trace` instead it could be used in pure setting as well.

io-sim/tags
```diff
@@ -0,0 +1,1930 @@
+!_TAG_FILE_SORTED	1	//
```
coot, line 1: I don't think you want to commit the `tags` file.

io-sim/test/Test/Control/Monad/IOSimPOR.hs
```diff
@@ -410,31 +414,64 @@ unit_issue113_nonDep =
...
+--
+-- If the test fails with "outer timeout" error it means that 'exploreSimTrace'
+-- is not lazy and there's something wrong with IOSimPOR implementation and
+-- the trace is being forced way too soon. If, on the other hand, this property
+-- fails with "inner timeout" error, it means that 'exploreSimTrace' is somewhat
+-- lazy, however the trace is not being lazily generated.
+--
+--
+prop_lazy :: Property
+prop_lazy =
```
coot, line 452:
```suggestion
prop_explore_endless_simulation :: Property
prop_explore_endless_simulation =
```

io-sim/test/Test/Control/Monad/IOSimPOR.hs
```diff
@@ -410,31 +414,64 @@ unit_issue113_nonDep =
...
+    counterexample "outer timeout"
+  $ within 15000000 -- 15 seconds
+  $ exploreSimTrace
+      id sim $ \_ trace -> do
+        let l = take 10
+              . traceEvents
+              $ trace
```
coot, line 459: Why not `takeUntil (\(t,_,_,_) -> t < timeLimit)` where `timeLimit` is some random (but positive & small) `Time` generated for this test.

coot (changes requested):

io-sim/tags
bolt12: You are right!

io-sim/test/Test/Control/Monad/IOSimPOR.hs
bolt12: That sounds good!

coot (approved):
LGTM

coot (approved):

io-sim/src/Data/List/Trace.hs
```diff
@@ -65,6 +69,32 @@ ppTrace :: (a -> String) -> (b -> String) -> Trace a b -> String
...
+take n _ | n <= 0 = Nil Nothing
+take _ (Nil a)    = Nil (Just a)
+take n (Cons b o) = Cons b (take (n - 1) o)
+
+-- | Take elements from the Trace while the predicate holds, converting each to ().
+takeWhile :: (b -> Bool) -> Trace a b -> Trace (Maybe a) b
+takeWhile _ (Nil a)          = Nil (Just a)
+takeWhile p (Cons b o)
+  | p b                       = Cons b (takeWhile p o)
+  | otherwise                 = Nil Nothing
```
coot, line 84:
```suggestion
takeWhile _ (Nil a)    = Nil (Just a)
takeWhile p (Cons b o)
  | p b                = Cons b (takeWhile p o)
  | otherwise          = Nil Nothing
```

PR #158, opened 2024-05-13
CI: Bump checkout to v4

nitinprakash96:
In my last PR, I saw a lot of warning because node16 is deprecated in actions. Hopefully it's okay to send a patch for this.
CleanShot 2024-05-13 at 17 41 20@2x


coot (approved):
Thanks.

Issue #154, opened 2024-05-08
`propExploration` test failure


bolt12:
```haskell
      propExploration:                         FAIL (743.31s)
        *** Failed! Falsified (after 52 tests and 76 shrinks):
        AreNotEqual
        Shrink2 {getShrink2 = Tasks [Task [WhenSet 0 0],Task [],Task [ThrowTo 1],Task [WhenSet 0 0,ThrowTo 1],Task [ThrowTo 1],Task [ThrowTo 1,WhenSet 1 1]]}
        Schedule control: ControlAwait [ScheduleMod (RacyThreadId [5],2) ControlDefault [(RacyThreadId [4],0),(RacyThreadId [4],1),(RacyThreadId [3],0),(RacyThreadId [3],1),(RacyThreadId [3],2),(RacyThreadId [2],0),(RacyThreadId [1],0),(RacyThreadId [1],1),(RacyThreadId [1],2),(RacyThreadId [4],2),(RacyThreadId [4],3)]]
        Thread {5} (5) delayed at time Time 0s
          until after:
            Thread {1}
            Thread {1} (1)
            Thread {2}
            Thread {3}
            Thread {3} (3)
            Thread {4}
            Thread {4} (4)

        (...)

        InternalError "assertion failure: Thread {4} not runnable"
        assertion failure: Thread {4} not runnable
        Use --quickcheck-replay=564358 to reproduce.
        Use -p '/propExploration/' to rerun this test only.
```

[full-log.txt](https://github.com/input-output-hk/io-sim/files/15249847/text.txt)

PR #159, opened 2024-05-16
Fix IOSimPOR test failure

bolt12:
we were missing updating the non-dependencies in the case where the thread is not concurrent. The thread is not concurrent so it makes sense to not discover any races but it still make sense to discover any non-dependencies to make sure we don't forget to include them as per our counterexample in #154 .

### DIFF
--- a/io-sim/src/Control/Monad/IOSimPOR/Internal.hs
+++ b/io-sim/src/Control/Monad/IOSimPOR/Internal.hs
@@ -1771,17 +1771,27 @@ updateRaces thread@Thread { threadId = tid }
                                            stepInfoRaces  } =
           -- if this step depends on the previous step, or is not concurrent,
           -- then any threads that it wakes up become non-concurrent also.
-          let !lessConcurrent = concurrent Set.\\ effectWakeup newEffect in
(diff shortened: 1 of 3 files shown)


bolt12:
Thank you @coot ! Yes you are right

io-sim/src/Control/Monad/IOSimPOR/Internal.hs
```diff
@@ -1771,17 +1771,27 @@ updateRaces thread@Thread { threadId = tid }
...
+                | happensBefore =           stepInfoNonDep
+                -- `newStep` did not happen after `step`
+                | otherwise     = newStep : stepInfoNonDep in
 
           if tid `notElem` concurrent
-            then stepInfo { stepInfoConcurrent = lessConcurrent }
+            then let
+                  in stepInfo { stepInfoConcurrent = lessConcurrent
+                              , stepInfoNonDep = stepInfoNonDep'
+                              }
```
coot, line 1790:
```suggestion
                  in stepInfo { stepInfoConcurrent = lessConcurrent
                              , stepInfoNonDep     = stepInfoNonDep'
                              }
```

coot (approved):
LGTM

PR #174, opened 2024-10-02
Attach label to TVars mentioned in Effects

jasagredo:
This ensures that `TxCommitted` events will also print labels in the `Effect`. Otherwise we get something like

```
TxCommitted [Labelled TVarId 0 SharedState] ... Effect { writes = [TVarId 0] }
```

now it will print:
```
TxCommitted [Labelled TVarId 0 SharedState] ... Effect { writes = [Labelled TVarId 0 SharedState] }
```

### DIFF
--- a/io-sim/src/Control/Monad/IOSimPOR/Internal.hs
+++ b/io-sim/src/Control/Monad/IOSimPOR/Internal.hs
@@ -568,10 +568,11 @@ schedule thread@Thread{
     CancelTimeout (Timeout tvar tmid) k -> do
       let timers' = PSQ.delete tmid timers
       written <- execAtomically' (runSTM $ writeTVar tvar TimeoutCancelled)
+      written' <- mapM someTVarToLabelled written
       (wakeup, wokeby) <- threadsUnblockedByWrites written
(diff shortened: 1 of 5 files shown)


io-sim/src/Control/Monad/IOSimPOR/Internal.hs
```diff
@@ -570,8 +570,9 @@ schedule thread@Thread{
       written <- execAtomically' (runSTM $ writeTVar tvar TimeoutCancelled)
       (wakeup, wokeby) <- threadsUnblockedByWrites written
       mapM_ (\(SomeTVar var) -> unblockAllThreadsFromTVar var) written
+      written' <- mapM someTVarToLabelled written
```
bolt12, line 573: could you perhaps move this above to line 570?

bolt12 (approved):

io-sim/src/Control/Monad/IOSimPOR/Internal.hs
jasagredo: Done

bolt12 (approved):

PR #176, opened 2024-10-02
Implement `MonadLabelledMVar`

jasagredo:


bolt12 (approved):

bolt12 (approved):

io-classes/src/Control/Concurrent/Class/MonadMVar.hs
```diff
@@ -204,6 +205,18 @@ instance MonadInspectMVar IO where
...
 
+-- | Labelled `MVar`s
+--
+-- The `IO` instances is no-op, the `IOSim` instance enhances simulation trace.
+-- This is very useful when analysing low lever concurrency issues (e.g.
+-- deadlocks, livelocks etc).
+class MonadMVar m
+   => MonadLabelledMVar m where
+  -- | Name an `MVar`
+  labelMVar         :: MVar m a -> String -> m ()
```
coot, line 216:
```suggestion
  labelMVar :: MVar m a -> String -> m ()
```

io-sim/src/Control/Monad/IOSim/STM.hs
```diff
@@ -261,13 +261,17 @@ data MVarState m a = MVarEmpty   !(Deque (TVar m (Maybe a))) -- blocked on take
 newEmptyMVarDefault :: MonadSTM m => m (MVarDefault m a)
 newEmptyMVarDefault = MVar <$> newTVarIO (MVarEmpty mempty mempty)
 
+labelMVarDefault
+  :: MonadLabelledSTM m
+  => MVarDefault m a -> String -> m ()
+labelMVarDefault (MVar tvar) = atomically . labelTVar tvar . (<> "-MVar")
```
coot, line 267: For the sake of consistency with other `label*` functions:
```suggestion
labelMVarDefault (MVar tvar) = atomically . labelTVar tvar
```

coot (changes requested):
Thanks @jasagredo, just minor comments.

io-sim/src/Control/Monad/IOSim/STM.hs
jasagredo: I added this because otherwise it is not possible to tell if a labelled TVar is actually an MVar. But I'm fine with removing it otherwise

coot: I see, that's fine.

io-classes/src/Control/Concurrent/Class/MonadMVar.hs
jasagredo: Done

coot (approved):
LGTM

PR #179, opened 2024-10-08
WIP: QSM

jasagredo:


io-sim/qsm/Test/StateMachine/IOSim/Execute.hs
```diff
@@ -0,0 +1,740 @@
...
+         in
+            ( prevProp .&&. pPre .&&. prop .&&. pPost .&&. pInv .&&. pMock
+            , prevCmds'
+            , insertConcretes vars cvars env
+            , transition smodel scmd sresp
+            , counter'
+            , transition cmodel ccmd cresp
+            )
+
+hasDoneWork :: (Eq p) => p -> [(a, p, c, SimEventType)] -> Bool
```
coot, line 605: Would this work? or maybe `ThreadId m` with a constraint if one needs a polymorphic function.
```suggestion
hasDoneWork :: IOSimThreadId -> [(a, IOSimThreadId, c, SimEventType)] -> Bool
```

io-sim/qsm-test/BasicTest/Correct.hs
```diff
@@ -0,0 +1,54 @@
...
+    return (AtomicCounter ref)
+
+incr :: MonadSTM m => AtomicCounter m -> m ()
+incr (AtomicCounter ref) =
+  atomically $ do
+    i <- readTVar ref
+    writeTVar ref (i + 1)
+
+get :: MonadSTM m => AtomicCounter m -> m Int
+get (AtomicCounter ref) = readTVarIO ref
```
jasagredo, line 31:
```suggestion
get (AtomicCounter ref) = do
  x <- readTVarIO ref
  traceSomething
  pure x
```

PR #188, opened 2024-10-15
Some tracing improvements

jasagredo:


io-classes/src/Control/Monad/Class/MonadSTM/Internal.hs
```diff
@@ -1257,3 +1261,40 @@ instance MonadSTM m => MonadSTM (ReaderT r m) where
...
+
+
+-- | Apply @f@ with the content of @tv@ as state, restoring the original value when an
+-- exception occurs
+withTMVar ::
+     (MonadSTM m, MonadThrow.MonadCatch m)
+  => TMVar m a
+  -> (a -> m (c, a))
+  -> m c
+withTMVar tv f = withTMVarAnd tv (const $ pure ()) (\a -> const $ f a)
```
amesgen, line 1323: Nitpick: Shouldn't this be called `modifyTMVar`? From `base`:
```haskell
withMVar :: MVar a -> (a -> IO b) -> IO b
modifyMVar :: MVar a -> (a -> IO (a, b)) -> IO b
```
(also, the order of arguments in the tuple should be flipped :sweat_smile:)

coot: The [`stm`](https://hackage.haskell.org/package/stm-2.5.3.1/docs/Control-Concurrent-STM-TMVar.html) package doesn't expose `withTMVar` that's why we don't expose it too.

io-classes/src/Control/Monad/Class/MonadSTM/Internal.hs
```diff
@@ -535,6 +544,51 @@ class MonadInspectSTM m
                       -> m ()
   traceTSemIO = \v f -> atomically (traceTSem Proxy v f)
 
+traceTVarShow :: (MonadTraceSTM m, Show a)
+               => proxy m
+               -> TVar m a
+               -> STM m ()
+traceTVarShow p tvar =
+  traceTVar p tvar (\pv v -> pure $ TraceString $ case (pv, v) of
+          (Nothing, st') -> "Created: " <> show st'
```
coot, line 553: This will never show up in the trace because when one calls `traceTVarShow`, the `tvar` already exists.

io-classes/src/Control/Monad/Class/MonadSTM/Internal.hs
```diff
@@ -535,6 +544,51 @@ class MonadInspectSTM m
...
+                 -> m ()
+traceTMVarShowIO tmvar =
+  traceTMVarIO tmvar (\pv v -> pure $ TraceString $ case (pv, v) of
+          (Nothing, Nothing) -> "Created empty"
+          (Nothing, Just st') -> "Created full: " <> show st'
+          (Just Nothing, Just st') -> "Put: " <> show st'
+          (Just Nothing, Nothing) -> "Remains empty"
+          (Just Just{}, Nothing) -> "Take"
+          (Just (Just st'), Just st'') -> "Modified: " <> show st' <> " -> " <> show st''
+      )
```
coot, line 591: Maybe a better name for these functions would be `debugTVar,debugTVarIO`?

io-classes/src/Control/Monad/Class/MonadSTM/Internal.hs
```diff
@@ -1257,3 +1311,40 @@ instance MonadSTM m => MonadSTM (ReaderT r m) where
...
+    )
+    (\(origState, _) -> \case
+        MonadThrow.ExitCaseSuccess (_, newState)
+          -> atomically $ putTMVar tv newState
+        MonadThrow.ExitCaseException _
+          -> atomically $ putTMVar tv origState
+        MonadThrow.ExitCaseAbort
+          -> atomically $ putTMVar tv origState
+    )
+    (uncurry f)
```
coot, line 1350: We don't include custom APIs like that, we restrain to what `stm` provides.

coot (changes requested):

io-classes/src/Control/Monad/Class/MonadSTM/Internal.hs
jasagredo: I suspected that, but still the type of the function made me pattern match on that. Should I throw an error instead?

io-classes/src/Control/Monad/Class/MonadSTM/Internal.hs
jasagredo: Sure thing!

io-classes/src/Control/Monad/Class/MonadSTM/Internal.hs
jasagredo: This function is now gone

io-classes/src/Control/Monad/Class/MonadSTM/Internal.hs
jasagredo: This function is now gone

io-classes/src/Control/Monad/Class/MonadSTM/Internal.hs
jasagredo: Done

coot (approved):
lgtm

PR #193, opened 2024-11-04
Use IntPSQ instead of OrdPSQ for `timers` map.

Saizan:
Using `IntPSQ` is much faster (20x) on some larger loads, and is recommended by `psqueues` where possible.

### DIFF
--- a/io-sim/CHANGELOG.md
+++ b/io-sim/CHANGELOG.md
@@ -12,6 +12,8 @@
   `TMVar`, `MVar` and a few others - except for `TChan`.
 - A blocked `takeTVar` is now safe in the presence of exceptions. It will relay
   the value to other waiting threads.
+- Faster handling of timeouts and timers by using a more efficient
+  internal representation.
 
 ## 1.6.0.0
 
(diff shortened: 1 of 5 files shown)


bolt12:
Great! Could you make the changes for IOSimPOR too?

Saizan:
@bolt12 sure! Btw, is there something we could use to benchmark the `runqueue` field? We likely want to change that to `HashPSQ`.

bolt12:
> @bolt12 sure! Btw, is there something we could use to benchmark the `runqueue` field? We likely want to change that to `HashPSQ`.

I guess you can use https://github.com/IntersectMBO/ouroboros-network/blob/main/ouroboros-network/sim-tests-lib/Test/Ouroboros/Network/Testnet.hs#L121

Saizan:
The tests performance seem unaffected by the changes, probably not enough threads to make it relevant, but should likely help otherwise.

Also I had to add an orphan `Hashable (Down a)` instance, let me know if you'd prefer it to be moved or if you have other comments about coding style.

bolt12 (approved):
LGTM! Thank you !!

PR #196, opened 2024-12-07
`tryReadTBQueueDefault`: fix bug when returning `Nothing`

### LINKED ISSUE #195: `tryReadTBQueue` is buggy in `IOSim`
I would expect the following program to print `0` twice:
```haskell
import Control.Concurrent.Class.MonadSTM
import Control.Monad.IOSim
import Numeric.Natural

main :: IO ()
main = do
  print =<< repro
  print $ runSimOrThrow repro

repro :: (MonadSTM m) => m Natural
repro = atomically $ do
  q <- newTBQueue 10
  _ <- tryReadTBQueue q
  writeTBQueue q ()
  _ <- tryReadTBQueue q
  lengthTBQueue q
```
However, it only works for `IO`; for `IOSim` is throws an exception:
```
0
*** Exception: FailureEvaluation arithmetic underflow
```
So something about `tryReadTBQueueDefault` in `IOSim` is incorrect.

amesgen:
Closes #195 

AFAICT, there is no need to update the state of the `TBQueue` when we didn't read anything.

### DIFF
--- a/io-sim/CHANGELOG.md
+++ b/io-sim/CHANGELOG.md
@@ -23,6 +23,7 @@
   is more general.  These functions now accepts trace with any result, rather
   than one that finishes with `SimResult`.
 - More polymorphic `ppTrace_` type signature.
+- Fixed `tryReadTBQueue` when returning `Nothing`.
 
 ## 1.6.0.0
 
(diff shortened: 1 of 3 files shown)


coot (approved):
Thanks Alex, LGTM

coot:
If you need it in `1.5.x` we can release `1.5.1.1` version.

amesgen:
> If you need it in `1.5.x` we can release `1.5.1.1` version.

I think it would be neat if it is not too much work, but it is not urgent for IOG stuff as one can work around this (I am also out of office until next year).

amesgen:
Ah, just realized I have merge rights here...

amesgen:
The CI failure seems unrelated to this PR

PR #201, opened 2024-12-19
Export `Time` from `Control.Monad.IOSim`

### LINKED ISSUE #200: io-sim: Export `Time` from `Control.Monad.IOSim`
**Is your feature request related to a problem? Please describe.**
Many types exposed from the `Control.Monad.IOSim` module have fields  of type `Time`. For example, the field `seTime` of `SimEvent`. However, the type `Time` is not exported from `Control.Monad.IOSim` and is only available from the deprecated `si-timers` package. The deprecated `si-timers` package recommends using the nearly equivalent modules from the `io-classses` package, but, unfortunately, the `io-classes` package does not export the `Time` type. Therefore, in order to use the `seTime` field of a `SimEvent`, one has to depend on the deprecated `si-timers` package for—and only for—that newtype.

**Describe the solution you'd like**
Export the `Time` newtype with its constructor from the `Control.Monad.IOSim` module. This isn't formally a breaking change, but _can be_ a breaking change if users are importing both `Control.Monad.IOSim` from `io-sim` and `Control.Monad.Class.MonadTime.SI` from `si-timers` unqualified. However, it should not be a breaking change in any codebase with import linting.

**Describe alternatives you've considered**
There are two alternative solutions:

- Unpack the `seTime` field and others to expose the underlying `DiffTime` value instead. This is the right solution, but is certainly a breaking change, which feels too major a change for such a minor improvement.
- Define a pattern synonym for the `SimEvent` constructor that unpacks `seTime` as `seDiffTime`. This isn't formally a breaking change, but _can be_ a breaking change if the pattern synonym name clashes with any common names, which should be easy enough to avoid. However, this feels a bit heavy handed for such a small change, as it would require numerous pattern synonyms.

**Are you willing to implement it?**
* [X] Yes.

wenkokke:
Fixes #200.

### DIFF
--- a/io-sim/CHANGELOG.md
+++ b/io-sim/CHANGELOG.md
@@ -22,6 +22,7 @@
   - `selectTraceEventsSayWithTime'`
   is more general.  These functions now accepts trace with any result, rather
   than one that finishes with `SimResult`.
+- Export `Time` from `Control.Monad.IOSim`.
 
 ## 1.6.0.0
 
--- a/io-sim/src/Control/Monad/IOSim.hs
+++ b/io-sim/src/Control/Monad/IOSim.hs
@@ -42,6 +42,7 @@ module Control.Monad.IOSim
   , unshareClock
     -- * Simulation trace
   , type SimTrace
+  , Time (..)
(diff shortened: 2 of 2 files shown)


bolt12 (approved):

wenkokke:
@bolt12 Should be ready to merge now.

PR #206, opened 2025-01-14
Support ghc 9.12

erikd:


io-classes/src/Control/Monad/Class/MonadSTM/Internal.hs
```diff
@@ -1030,7 +1030,6 @@ unGetTBQueueDefault (TBQueue rsize read wsize _write _size) a = do
 -- | Default implementation of 'TArray'.
 --
 data TArrayDefault m i e = TArray (Array i (TVar m e))
-  deriving Typeable
```
erikd, line 1033: GHC as derived `Typeable` by default since `ghc-8.6` or something. With `ghc-9.12` a warning was added for unneeded derived instances.

coot (approved):
LGTM

PR #212, opened 2025-05-07
Provide MonadTraceMVar

### LINKED ISSUE #18: Provide MonadTraceMVar
First provide `MonadInspectMVar` which is the analogy of `MonadInspectSTM` but for `MVar`s.
```haskell
class ( MonadMVar m
      , Monad (InspectMonadMVar m)
      )
    => MonadInspectMVar m where
    type InspectMonadMVar m :: Type -> Type
    inspectMVar :: proxy m -> MVar m a -> InspectMonadMVar m (Maybe a)
```

Then provide a type class
```haskell
class MonadTraceMVar m where
  traceMVarIO :: proxy 
              -> MVar m a
              -> (Maybe (Maybe a) -> Maybe a -> InspectMonad m TraceValue)
              -> m ()
```

and two instances: `MonadTraceMVar IO` and `MonadTraceMVar (IOSim s)`.

Under this proposal, I also suggest we also rename the associated type family `InspectMonad` of `MonadInspectSTM` as `InsepctMonadSTM`.

bolt12:
Closes #18

### DIFF
--- a/io-sim/src/Control/Monad/IOSim/Types.hs
+++ b/io-sim/src/Control/Monad/IOSim/Types.hs
@@ -575,7 +575,7 @@ instance MonadSTM (IOSim s) where
   cloneTChan        = MonadSTM.cloneTChanDefault
 
 instance MonadInspectSTM (IOSim s) where
-  type InspectMonad (IOSim s) = ST s
+  type InspectMonadSTM (IOSim s) = ST s
   inspectTVar  _                 TVar { tvarCurrent }  = readSTRef tvarCurrent
   inspectTMVar _ (MonadSTM.TMVar TVar { tvarCurrent }) = readSTRef tvarCurrent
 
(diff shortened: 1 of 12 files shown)


io-sim/src/Control/Monad/IOSim/Types.hs
```diff
@@ -615,6 +615,16 @@ instance MonadInspectMVar (IOSim s) where
         MVarEmpty _ _ -> pure Nothing
         MVarFull x _  -> pure (Just x)
 
+instance MonadTraceMVar (IOSim s) where
+  traceMVarIO _ (MVar mvar) f = traceTVarIO mvar traceMVarAsTVar
```
coot, line 619: The drawback of it is that everything is traced using `TraceTVar` constructor.  Could you add `TraceMVar` to `StmA` and use it here instead?

coot (changes requested):

io-sim/src/Control/Monad/IOSim/Types.hs
coot: Ignore my comment; `TraceTVar` constructor is only used to write to the `TVar` callback.

coot (approved):
LGTM

PR #213, opened 2025-05-15
Release io-classes-1.8

coot:
- **Reworded CHANGELOG.md**
- **io-classes: prefixed sublibraries**
- **Bump versions to 1.8.0.0**

### DIFF
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ghc: ["8.10", "9.2", "9.4", "9.6", "9.8", "9.10", "9.12"]
+        ghc: ["9.6", "9.8", "9.10", "9.12"]
         os: [ubuntu-latest, macos-latest, windows-latest]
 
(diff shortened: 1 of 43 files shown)


jorisdral:
Hiya, I wanted to ask why the sub-libraries are renamed? IIUC, there's no way in cabal to say "depend on `io-classes:strict-mvar ^>= 1.7` *or* depend on `io-classes:io-classes-strict-mvar ^>=1.8`". So changing a sub-library name means a dependent package could never support both the old and new sublibrary. It's kind of similar re-uploading a package to Hackage under a new name, and deprecating the old one. 

Feel free to do with this info what you want of course -- maybe you've already considered what I said. I wanted to point it out just in case

coot:
With input-output-hk/haskell.nix#2363 we might not need the new prefixes.

coot:
The issue is fixed, I dropped the prefixes :smile:.

jasagredo (approved):
Except the `MonadTraceMVar (IOSim s)` instance, all changes are removing only Pragmas so it looks fine. The changes there also look reasonable.

PR #215, opened 2025-05-26
ghc-9.2 and ghc-9.10.2  support

coot:


coot:
@jasagredo could you review it?

PR #217, opened 2025-06-03
Support for `ghc-9.2` in `io-sim`

geo2a:
This is a follow-up to https://github.com/input-output-hk/io-sim/pull/215 motivated by the build failures in `lsm-tree` (see https://github.com/IntersectMBO/lsm-tree/actions/runs/15419010635/job/43388792198?pr=749)

### DIFF
--- a/io-sim/CHANGELOG.md
+++ b/io-sim/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 ### Non-breaking changes
 
+### 1.8.0.1
+
+* Added support for `ghc-9.2`.
+
 ## 1.8.0.0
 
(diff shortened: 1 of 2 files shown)


coot:
@jorisdral, @geo2a do you need support for `GHC-9.{2,4}`? I dropped both of them from CI some time ago...

geo2a:
> `io-sim-1.8.0.1` is on hackage
Thanks @coot! 
> do you need support for `GHC-9.{2,4}`
I think it would be nice to test with GHC 9.2 and 9.4 in CI if it is not too much trouble.

PR #223, opened 2025-07-29
Unique Symbol Generation

LSLeary:
The project I want to test with *io-sim* uses `Data.Unique`, so I wrote this.

P.s. I'm not yet familiar with the library, let alone its internals—do let me know if I've written something silly.


io-sim/src/Control/Monad/IOSim/CommonTypes.hs
```diff
@@ -28,6 +29,7 @@ module Control.Monad.IOSim.CommonTypes
   , BlockedReason (..)
   , Labelled (..)
   , ppLabelled
+  , IOSimUnique(..)
```
coot, line 32:
```suggestion
  , IOSimUnique
```

io-sim/src/Control/Monad/IOSim/CommonTypes.hs
```diff
@@ -202,6 +204,12 @@ ppLabelled :: (a -> String) -> Labelled a -> String
 ppLabelled pp Labelled { l_labelled = a, l_label = Nothing  } = pp a
 ppLabelled pp Labelled { l_labelled = a, l_label = Just lbl } = concat ["Labelled ", pp a, " ", lbl]
 
+-- | Abstract unique symbols à la "Data.Unique".
+type role IOSimUnique nominal
+newtype IOSimUnique s = MkUnique{ unMkUnique :: Integer }
+  deriving stock   (Eq, Ord)
+  deriving newtype (Hashable, NFData)
+
```
coot, line 212: It's easier to read when data is defined first.
```suggestion
newtype IOSimUnique s = MkUnique{ unMkUnique :: Integer }
  deriving stock   (Eq, Ord)
  deriving newtype (Hashable, NFData)
type role IOSimUnique nominal
```

io-sim/src/Control/Monad/IOSim/CommonTypes.hs
coot: `base` doesn't export the constructor, can we hide it too?

io-sim/src/Control/Monad/IOSimPOR/Internal.hs
```diff
@@ -814,6 +816,11 @@ schedule thread@Thread{
       let thread' = thread { threadControl = ThreadControl k ctl }
       schedule thread' simstate
 
+    NewUnique k -> do
+      let thread'   = thread{ threadControl = ThreadControl (k nextUniq) ctl }
+          simstate' = simstate{ nextUniq = MkUnique (unMkUnique nextUniq + 1) }
+      schedule thread' simstate'
```
coot, line 822: If we implement `IOSimUnique` using `STRef s` (or `TVar m`) then the `MonadUnique` instance will require `MonadST` (or `MonadSTM`), which would be a disadvantage for users.  This approach doesn't require it :+1:, but it needs two implementations (`IOSim` and `IOSimPOR`).

coot, line 822: We should add a trace event here for creating a new unique.

io-sim/src/Control/Monad/IOSim/Internal.hs
```diff
@@ -631,6 +633,11 @@ schedule !thread@Thread{
           thread' = thread { threadControl = ThreadControl k' ctl }
       schedule thread' simstate
 
+    NewUnique k -> do
+      let thread'   = thread{ threadControl = ThreadControl (k nextUniq) ctl }
+          simstate' = simstate{ nextUniq = MkUnique (unMkUnique nextUniq + 1) }
+      schedule thread' simstate'
```
coot, line 639: Please add a trace event.

io-classes/src/Control/Monad/Class/MonadUnique.hs
```diff
@@ -0,0 +1,33 @@
...
+
+instance MonadUnique IO where
+  type Unique IO = IO.Unique
+  newUnique  = IO.newUnique
+  hashUnique = IO.hashUnique
+
+instance MonadUnique m => MonadUnique (ReaderT r m) where
+  type Unique (ReaderT r m) = Unique m
+  newUnique  = lift newUnique
+  hashUnique = hashUnique @m
```
coot, line 32: Could you also add instances for other mtl type classes in `io-classes:mtl` sublibrary?

io-sim/src/Control/Monad/IOSim/CommonTypes.hs
LSLeary: It's properly hidden when re-exporting from `Control.Monad.IOSim`; I only expose the constructor internally to bump uniques in each `schedule`.

io-sim/src/Control/Monad/IOSim/CommonTypes.hs
LSLeary: Done.

io-sim/src/Control/Monad/IOSimPOR/Internal.hs
LSLeary: Done.

io-classes/src/Control/Monad/Class/MonadUnique.hs
LSLeary: Done.

io-sim/src/Control/Monad/IOSimPOR/Internal.hs
LSLeary: I don't really know the codebase well enough to weigh the merits of different paths; I just took the first one I found.

io-sim/src/Control/Monad/IOSim/CommonTypes.hs
coot: ok. small style change:
```suggestion
  , IOSimUnique (..)
```

coot (commented):
We can trace `Integer` rather than `IOSimUnique`, this will avoid the `s` parameter.

io-classes/src/Control/Monad/Class/MonadUnique.hs
```diff
@@ -0,0 +1,33 @@
...
+import Data.Unique qualified as IO
+
+-- transformers
+import Control.Monad.Reader (ReaderT(..), lift)
+
+
+class (Monad m, Eq (Unique m), Ord (Unique m)) => MonadUnique m where
+  type Unique m :: Type
+  newUnique  :: m (Unique m)
+  hashUnique :: Unique m -> Int
```
coot, line 22: The `hashUnique` cannot be used in this form, because GHC has no way to find out `m` from `Unique m` (turning on `AllowAmbiguousTypes` was not a good idea, as in this case it indicates a real problem).

There are two ways of fixing this, either adding `Proxy` arguments, but that changes API, or using `TypeFamilyDependencies`.  In `io-classes` we prefer the latter (to stay as close to `base` API as possible), although it requires some newtype wrappers for instances (e.g. monad transformers).
```suggestion
class (Monad m, Eq (Unique m), Ord (Unique m)) => MonadUnique m where
  type Unique m = (unique :: Type) | unique -> m
  newUnique  :: m (Unique m)
  hashUnique :: Unique m -> Int
```

io-classes/src/Control/Monad/Class/MonadUnique.hs
```diff
@@ -0,0 +1,33 @@
+{-# LANGUAGE TypeFamilies, AllowAmbiguousTypes #-}
```
coot, line 1:
```suggestion
{-# LANGUAGE TypeFamilies           -#}
{-# LANGUAGE TypeFamilyDependencies -#}
```
Explained below.

io-classes/src/Control/Monad/Class/MonadUnique.hs
coot, line 32: You'll need `UniqueReadrT` newtype wrapper, check out the `MonadAsync` instances how it's done there.

io-sim/src/Control/Monad/IOSim/Types.hs
```diff
@@ -626,6 +629,11 @@ instance MonadTraceMVar (IOSim s) where
 instance MonadLabelledMVar (IOSim s) where
   labelMVar = labelMVarDefault
 
+instance MonadUnique (IOSim s) where
+  type Unique (IOSim s) = IOSimUnique s
+  newUnique  = IOSim (oneShot NewUnique)
+  hashUnique = hash
```
coot, line 635: `base` is using [`integerToInt`](https://hackage.haskell.org/package/ghc-internal-9.1201.0/docs/src/GHC.Internal.Data.Unique.html#hashUnique), let's do the same.

io-sim/src/Control/Monad/IOSim/CommonTypes.hs
LSLeary: Done.

io-classes/src/Control/Monad/Class/MonadUnique.hs
LSLeary: I had originally intended to mention/ask about this, but it slipped my mind. When choosing between an ambiguous type and an injective type family, I begrudgingly settled for the former to allow the `ReaderT` instance through, but I guess necessary type applications are as big of an API change as added proxies, so it's no good indeed. I don't know why the `newtype` injectivity fix didn't occur to me ... Done.

io-classes/src/Control/Monad/Class/MonadUnique.hs
LSLeary: `TypeFamilyDependencies` implies `TypeFamilies`; it's redundant here. But yeah, `AllowAmbiguousTypes` is gone now.

io-classes/src/Control/Monad/Class/MonadUnique.hs
LSLeary: Done.

io-sim/src/Control/Monad/IOSim/Types.hs
LSLeary: `IntegerToInt` is only exposed from `GHC.*` modules; it's not meant to be used directly. The standard way to use it would be `fromInteger`, which I mistakenly thought was all `hash @Integer` did. Fixed.

LSLeary: Oops, fixed this in the wrong place and left an inconsistent `Hashable` instance hanging. "Fix" fixed.

io-classes/src/Control/Monad/Class/MonadUnique.hs
coot: You're right, no need for both, but it's fine.

io-classes/src/Control/Monad/Class/MonadUnique.hs
coot: Thanks

io-classes/src/Control/Monad/Class/MonadUnique.hs
```diff
@@ -0,0 +1,48 @@
...
+
+instance MonadUnique IO where
+  type Unique IO = IO.Unique
+  newUnique  = IO.newUnique
+  hashUnique = IO.hashUnique
+
+
+newtype UniqueFor t m = MkUniqueFor{ unMkUniqueFor :: Unique m }
+deriving instance MonadUnique m => Eq  (UniqueFor r m)
+deriving instance MonadUnique m => Ord (UniqueFor r m)
```
coot, line 45: Clever :), maybe we should adopt this pattern in other places...

io-sim/src/Control/Monad/IOSim/Internal.hs
```diff
@@ -142,8 +142,9 @@ data SimState s a = SimState {
        timers   :: !(Timeouts s),
        -- | list of clocks
        clocks   :: !(Map ClockId UTCTime),
-       nextVid  :: !VarId,     -- ^ next unused 'VarId'
-       nextTmid :: !TimeoutId   -- ^ next unused 'TimeoutId'
+       nextVid  :: !VarId,          -- ^ next unused 'VarId'
+       nextTmid :: !TimeoutId,      -- ^ next unused 'TimeoutId'
+       nextUniq :: !(Unique s) -- ^ next unused @'Unique' s@
```
coot, line 147:
```suggestion
       nextVid  :: !VarId,     -- ^ next unused 'VarId'
       nextTmid :: !TimeoutId, -- ^ next unused 'TimeoutId'
       nextUniq :: !(Unique s) -- ^ next unused @'Unique' s@
```

io-sim/src/Control/Monad/IOSim/Internal.hs
```diff
@@ -631,6 +633,13 @@ schedule !thread@Thread{
           thread' = thread { threadControl = ThreadControl k' ctl }
       schedule thread' simstate
 
+    NewUnique k -> do
+      let thread'   = thread{ threadControl = ThreadControl (k nextUniq) ctl }
+          n         = unMkUnique nextUniq
+          simstate' = simstate{ nextUniq = MkUnique (n + 1) }
```
coot, line 639: if we had `Enum` instance on `Unique` type in `io-sim`:
```suggestion
          simstate' = simstate{ nextUniq = succ nextUniq }
```

io-sim/src/Control/Monad/IOSimPOR/Internal.hs
```diff
@@ -814,6 +816,13 @@ schedule thread@Thread{
       let thread' = thread { threadControl = ThreadControl k ctl }
       schedule thread' simstate
 
+    NewUnique k -> do
+      let thread'   = thread{ threadControl = ThreadControl (k nextUniq) ctl }
+          n         = unMkUnique nextUniq
+          simstate' = simstate{ nextUniq = MkUnique (n + 1) }
```
coot, line 822:
```suggestion
          simstate' = simstate{ nextUniq = succ nextUniq }
```

coot (commented):
Thank @LSLeary, the PR looks good to me.  Just some minor things.  Please squash the commits and I'll merge it.

io-sim/src/Control/Monad/IOSim/Internal.hs
LSLeary: `Unique` can't have `Enum`—if someone else can create your `Unique` by `succ`ing theirs, yours isn't unique.

PR #251, opened 2026-03-24
Add IOSimPOR QuickCheck monadic combinators

### LINKED ISSUE #229: Missing `PropertyM` callbacks for `IOSimPOR`
**Is your feature request related to a problem? Please describe.**
The `monadicIOSim` callback is great, it allows you to write nice properties on top of `PropertyM` for `IOSim`. However, it runs `runSimTraceST` and not the `IOSimPOR` scheduler (for obvious reasons).

When we want to do `IOSimPOR` testing we have to do some spicy stuff roughly like this:
```haskell
basicIOSimPOR :: Testable a => (forall s. PropertyM (IOSim s) a) -> Property
basicIOSimPOR prop = forAllBlind prop' $ \ p -> exploreSimTrace id p $ \ _ tr ->
    either (flip counterexample False . show) id $ traceResult False tr
  where
    prop' :: Gen (forall s. IOSim s Property)
    prop' = do
      Capture eval <- capture
      pure $ eval $ monadic' prop
```
As far as I can tell a version of this function could be in `io-sim` and currently isn't.

**Describe the solution you'd like**
A generalization of `exploreSimTrace` to `PropertyM`

**Describe alternatives you've considered**
Doing it by hand like above.

**Are you willing to implement it?**
* [x] Are you? :smiley:

dancewithheart:
Fixes #229

Add `IOSimPOR` QuickCheck monadic combinators:

- `monadicIOSimPOR`
- `monadicIOSimPOR_`
- `runIOSimPORGen`

similar to `monadicIOSim`, `monadicIOSim_`, and `runIOSimGen`.

This removes the need for hand-written `capture` + `exploreSimTrace`
plumbing when using `PropertyM` with `IOSimPOR`.

`runIOSimPORGen` keeps the callback shape of `exploreSimTrace`, rather
than `runIOSimGen`, to preserve the POR-specific callback power.

`monadicIOSimPOR_` provides a convenient default wrapper, similar to
`monadicIOSim_`.

### DIFF
--- a/io-sim/CHANGELOG.md
+++ b/io-sim/CHANGELOG.md
@@ -24,6 +24,8 @@
 * Added `Data.List.Trace.last`
 * Although `IOSim` and `IOSimPOR` are pure we use `evaluate` in a few places,
   non of them now catch asynchrounous exceptions.
+* Added `IOSimPOR` QuickCheck monadic combinators:
+  `monadicIOSimPOR`, `monadicIOSimPOR_` and `runIOSimPORGen`.
 
 ## 1.9.1.0
 
--- a/io-sim/src/Control/Monad/IOSim.hs
+++ b/io-sim/src/Control/Monad/IOSim.hs
@@ -18,6 +18,9 @@ module Control.Monad.IOSim
   , monadicIOSim
   , runIOSimGen
     -- ** Explore races using /IOSimPOR/
+  , monadicIOSimPOR_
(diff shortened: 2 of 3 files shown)


MaximilianAlgehed (approved):

coot (approved):
LGTM. Do you need a Hackage release?

